PR #7, opened 2017-03-28
New custom generator for PWGDQ-LMee: HF dielectron triggered Pythia

miweberSMI:


preghenella:
I tested the pull request and it works.
Waiting for comments from PWG conveners before merging to the master.

iarsene:
+1

preghenella:
I'm not sure I understand the comment from @iarsene, but I take it is and positive one.

iarsene:
Hi Roberto, you refer to my "+1" ? I also commented on the corresponding JIRA ticket. You said there that the PWG conveners should approve the pull request. So i thought that "+1" means i approve. I only forked today AliDPG and I do not actually know what is the chain of approvals here... whether its the same as in AliPhysics ...

preghenella:
yes, sorry I missed the comment on JIRA.
There is still not an official chain, at least not until everything will be in place and properly advertised. Thanks!

iarsene:
Ok, then I should just comment in words on the pull requests, and let you do the "+1" and "+test"

preghenella:
Soon we will have something like AliRoot and you will be able to do it yourself :-)

Roberto Preghenella
http://about.me/preghenella

PR #14, opened 2017-04-12
Added custom generator for diffraction tuning of Pythia

valzacc:


valzacc:
Ciao Roberto,
You are right. Keep only the lines, which do not overlap (i.e. the ones you mentioned).

Regarding the crossing angle, I was also wondering about it.
Can you please tell me why and when we moved from setting 0.00028 to 0.? 

Thanks!
Valentina

preghenella:
please updated the pull request keeping only the needed lines.

Regarding the cross-angle, it was changed to zero following the suggestion of Chiara Oppedisano.
Now I don't remember what was the reasoning, you might get in touch with her for clarifications.

valzacc:
Is it permanently 0. now for all the regular productions?

preghenella:
ps: also notice that the PHOJET configuration is not passing the tests.

The pull request will not be accepted until this is fixed and I am not 100% sure what the issue is.
Plain PHOJET runs fine, therefore I guess that the 'floating point exception' is triggered by either the 'crossing-angle' or the 'diff-tune' settings.

I unfortunately don't have a way to fix it quickly, as I

valzacc:
Ok, let's do it like this: I will keep the crossing angle 0. default setting now, and I will speak with Chiara O. later, so that we can see if Phojet fails because of this. If that's not the reason, the problem is bigger since the diffraction tuning is done in AliRoot and I need to look at it there. 

I close now this pull request and open a new one with these small fixes, if Phojet fails let me know there.

preghenella:
> Is it permanently 0. now for all the regular productions?
yes

preghenella:
> Ok, let's do it like this: I will keep the crossing angle 0. default setting now, and I will speak with Chiara O. later, so that we can see if Phojet fails because of this. If that's not the reason, the problem is bigger since the diffraction tuning is done in AliRoot and I need to look at it there.

I close now this pull request and open a new one with these small fixes, if Phojet fails let me know there.

I will check by myself if without crossing angle PHOJET survives. I get back to you soon

preghenella:
ps: do not close this pull request, just updated it.
I think it should be enough that you update your master...

valzacc:
yes you are right, I have reopened it and new commits are taken from my fork. Thanks

preghenella:
PHOJET still crashes, the issue is in the DiffTune.
I will open a JIRA ticket for that.

valzacc:
Thanks, please add me as a watcher there!

valzacc:
Let me do it

PR #21, opened 2017-04-21
Added Str002 MC generator: strangness injection with reduced range

ddobrigk:


preghenella:
I'm not going to test it because I guess you copy-paste the other file and carefully edited what was needed.

PR #26, opened 2017-05-16
run the MUON TRK efficiency in only one group

pillot:


chiarazampolli:
Ciao Philippe,

Sorry for the late reply. I see that you added a wagon to a group. Did you check that the memory stays under control? I guess you don't use trees nor ntuples, not THnSparses.

Chiara

pillot:
Ciao Chiara,

The wagon was already there, but active in all groups by default, so are getting 5 times the stat after the merging. I disable it in all groups but one, so in term of memory it should be even better ;-)

Cheers,
Philippe

chiarazampolli:
Hello Philippe,
Thanks! Would you like this commit in the tag for 2017 data?
Cheers,
Chiara

pillot:
Thanks Chiara!
Yes, it would be nice if it is not too late.
Cheers
Philippe

chiarazampolli:
Ciao,
I think we should be on time - depending on Darios availability to make the tag.

Cheers,

Chiara

PR #29, opened 2017-05-24
Structured parameter options (Christoph) and fix in rapidity bins

ekryshen:
More from Christoph: new flat generator GeneratorFlat2Trk and new
processes (e. g. kCohJpsiToProton)

### DIFF
--- a/DataProc/CPass1/recCPass1.C
+++ b/DataProc/CPass1/recCPass1.C
@@ -79,7 +79,7 @@ void recCPass1(const char *filename="raw.root",Int_t nevents=-1, const char *ocd
   }
 
   // All friends
-  rec.SetFractionFriends(2.0);
+  rec.SetFractionFriends(.2);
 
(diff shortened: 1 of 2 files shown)


ekryshen:
I agree. It might be better to introduce this generator as a separate instance in GeneratorConfig.C instead of custom generator. I will remove that part for the moment (not needed for the pending production) and we will insert it later.

preghenella:
For some reason, the comment I posted yesterday evening did not post here properly.
In this pull request there is an extra commit
	Decreased esd friends for 2017 raw data reconstruction			3d2c462
which should not be there.
This commit is already in the master. @ekryshen, you should update your fork for the pull request to be properly propagated.

Anyway, I will take care of it by cherry-picking only the relevant commits
	Structured parameter options (Christoph) and fix in rapidity bins  …			f0201fa
	GeneratorFlat2Trk removed for the moment			0d50713

Although, I will be unable to test the changes.

PR #30, opened 2017-05-30
Nuclei generator for pp collisions

mpuccio:
Hi @preghenella,

these are the changes and the custom generator required by [ALIROOT-7306](https://alice.its.cern.ch/jira/browse/ALIROOT-7306).
I already tested but please have a look and tell me if you find something pathological.

Thanks a lot!
Cheers,
Max

### DIFF
--- a/MC/CustomGenerators/PWGLF/Pythia8_Monash2013_Nuclex001.C
+++ b/MC/CustomGenerators/PWGLF/Pythia8_Monash2013_Nuclex001.C
@@ -1,0 +1,14 @@
+AliGenerator *
+GeneratorCustom()
+{
+  AliGenCocktail *ctl = GeneratorCocktail("Monash2013_Nuclex001");
+  // pythia8
+  AliGenerator   *py8 = GeneratorPythia8(kPythia8Tune_Monash2013);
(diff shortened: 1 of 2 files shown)


preghenella:
hi @mpuccio,
not important, though: is this config the same as PWGLF/Pythia6_Perugia2011_Nuclex001.C in terms of nuclei injection?

mpuccio:
Hi @preghenella,

no: we decided to reduce the number of injected particle and to remove the species that are very unlikely to be seen with the available statistics (e.g. 4He and hyper-triton).
Maybe I can rename it to Nuclei instead of Nuclex to remark that the injection scheme is different.
If it is fine with you I can do that.

Cheers,
Max

preghenella:
I am not very much concerned about the names, but rather on the continuous proliferation of different configurations that are supposed to deal with the same physics needs of a given PWG-PAG.

Anyway, as the above is none of my business, I'm fine with it.

mpuccio:
You are right. Unfortunately we are starting to optimise a bit the injection schemes only now for lack of time. We will improve on this.

preghenella:
there is not problem for that: you'll keep track of your configurations.
I'm testing, if successful I will merge the pull request.

PR #31, opened 2017-06-01
Add CVMFS OCDB support in simulation

dberzano:
Along with `--ocdb [alien|local]` the `cvmfs[:/full/path]` is now supported. The
full path is optional and defaults to `/cvmfs/alice-ocdb.cern.ch`.

AliRoot version must be compatible with it.

### DIFF
--- a/AOD/AODtrainsim.C
+++ b/AOD/AODtrainsim.C
@@ -108,7 +108,7 @@ void AODtrainsim(Int_t merge=0)
     gROOT->LoadMacro("$ALIDPG_ROOT/MC/OCDBConfig.C");
     OCDBDefault(1);    
   }
-  else if (ocdbConfig.Contains("alien")) {
+  else if (ocdbConfig.Contains("alien") || ocdbConfig.Contains("cvmfs")) {
     // set OCDB 
     gROOT->LoadMacro("$ALIDPG_ROOT/MC/OCDBConfig.C");
     OCDBDefault(1);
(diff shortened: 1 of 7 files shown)


dberzano:
Hi @preghenella, this is the work done by @mconcas to add CVMFS OCDB support for the MC validation. Current use of `dpgsim.sh` (with `--ocdb` different from  `cvmfs`) does not break existing use cases, so it should be safe to merge. Can you please have a look?

preghenella:
sure, I will have a look asap.
Thanks

dberzano:
We are adding one more small thing before merging, hold on.

dberzano:
Actually this is ready to be merged, we are not adding anything new. It is however important IMHO to refactor a small part of `AOD/AODtrainsim.C` around the line 127.

The variable `doCDBconnect` suggests that, if `kTRUE`, then a `TGrid::Connect()` is performed. However the connect is done _either_ if this variable's true, _or_ if we are currently at a merging stage.

The only case where `doCDBconnect` is set to `kFALSE` is in `isMuonCaloPass` (see `UpdateFlags()`).

I think we have to drop the very confusing `doCDBconnect` variable (since a connection might occur even if it's false) and use `isMuonCaloPass` directly for the condition at line 127.

@preghenella: what do you think?

preghenella:
I would take care of the OCDBconnect in another ticket.
If it is indeed useless, we can get rid of it.

dberzano:
Many thanks!

PR #32, opened 2017-06-12
new settings for muon only MC

pillot:
Hi,

I added a new settings to run muon-only MC, like the ones we use for tracking systematics studies and for J/Psi, Upsilon, etc... AccEff correction calculations. This settings is different from the "muon" settings that correspond to the muon_calo passes and can be used for e.g. tracklet multiplicity studies or by UPC for STARLIGHT MC.

I tested successfully this new settings locally with single muon MC.

Also I tried to be careful not to change any existing behaviour with those modifications, but a double check is more than welcome! :-)

Cheers,
Philippe

### DIFF
--- a/AOD/AODtrainsim.C
+++ b/AOD/AODtrainsim.C
@@ -22,6 +22,7 @@ Bool_t      useCORRFW           = kFALSE;  // do not change
 Bool_t      useAODTAGS          = kFALSE;  // use AOD tags
 Bool_t      useSysInfo          = kFALSE;  // use sys info
 Bool_t      isMuonCaloPass      = kFALSE;  // setting this to kTRUE will disable some not needed analysis tasks for a muon_calo pass
+Bool_t      isMuonOnly          = kFALSE;  // setting this to kTRUE will disable same tasks as isMuonCaloPass and some more
 
 // ### Analysis modules to be included. Some may not be yet fully implemented.
 //==============================================================================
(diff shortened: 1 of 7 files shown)


preghenella:
Hi @pillot, I looked at your changes and they look ok to me.
One question: are they all backward compatible wrt. AliRoot/AliPhysics?

What I mean is whether there is any piece of code that you added that works only for recent AliRoot/AliPhyisics. For instance, I noticed
{code}
-      AliAnalysisTaskTrigChEff *taskMuonTrig = AddTaskMTRchamberEfficiency();
 +      AliAnalysisTaskTrigChEff *taskMuonTrig = AddTaskMTRchamberEfficiency(isMuonOnly);
{code}
where before we called the AddTask without parameters, whereas now we call if with a parameter.
My question is whether this change is supported also in old versions of the task. This look to me to be ok, at least up to a few AliPhysics releases back.

preghenella:
another comment, @pillot.
I would have written the part in dpgsim.sh that does the automating setting from CONFIG_MODE
`### automatic settings from CONFIG_MODE`
in a way that it's easier for the reader to grasp what is going on. The way you did is ok, but I would like to ask you to please modify to something like

`
if [[ $CONFIG_MODE == *"MuonOnly"* ]]; then
### do settings for MuonOnly case
elif if [[ $CONFIG_MODE == *"Muon"* ]]; then
### do setting for Muon case
### (even is it is *Muon* we excluse *MuonOnly* as it is covered before)
fi
`

Thanks

pillot:
Hi @preghenella 

Thanks for the code review. Indeed the settings were not very clear. I made the modifications you suggested.

About the backward compatibility with old versions of AliRoot/AliPhysics, the changes in AliDPG did not requested any modification there. In the particular case of the code you mentioned, it has not been touched since 2013. Just the default parameter is not suited for that particular settings.

Cheers,
Philippe

PR #34, opened 2017-06-21
GammaJet and GammaTriggeredJet configurations for PYTHIA8

miweberSMI:
Dear Roberto,

here are the changes we discussed, can you have a look and approve if OK with you?

Thanks,
Michael

### DIFF
--- a/MC/GeneratorConfig.C
+++ b/MC/GeneratorConfig.C
@@ -17,7 +17,11 @@ enum EGenerator_t {
   // Pythia8
   kGeneratorPythia8, kGeneratorPythia8_Monash2013,
   // Pythia8 jets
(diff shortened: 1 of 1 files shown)


preghenella:
Thanks, @miweberSMI,
looks ok to me.
I merge it to the master.

There is one thing that I remember from one comment from Gustavo, he wanted us to remove the line
`pythia->SetFragPhotonInBarrel (kTRUE); // it will be driven mostly by decays, but should not harm`
and similar ones from the Gamma-Triggered configurations.

There must be a comment on JIRA.
Better to take care of that with another commit, if you don't mind.

Closing here

preghenella:
ps: as it seems you still don't have commit rights, I will take care of porting to the production branches

PR #35, opened 2017-06-21
configuration for Pythia6/8 events with gamma/jet-jet events or jet-j…

gconesab:
…et with gamma trigger in different gamma acceptances


preghenella:
thanks, @gconesab, looks ok.

PR #39, opened 2017-07-21
Added custom DIME generator

ekryshen:
This commit is related to this production request ALIROOT-7398

### DIFF
--- a/MC/CustomGenerators/PWGUD/Dime.C
+++ b/MC/CustomGenerators/PWGUD/Dime.C
@@ -1,20 +1,26 @@
 AliGenerator* GeneratorCustom() {
   printf("Starting DIME generator\n");
+  gSystem->Load("libEVGEN");
+  gSystem->Load("libTDime");
+  gSystem->Load("libdime");
 
   TString proc;
   if      (processConfig.Contains("Pipm")) proc = "pipm";
   else if (processConfig.Contains("Kpkm")) proc = "kpkm";
-  else printf("ERROR: uknown process\n");
+  else {
+    printf("ERROR: unknown process\n");
+    return 0;
+  }
 
   TString ff;
   if      (processConfig.Contains("Orexp")) ff = "orexp";
   else if (processConfig.Contains("Exp"))   ff = "exp";
   else if (processConfig.Contains("Power")) ff = "power";
-  else printf("ERROR: uknown process\n");
+  else {
+    printf("ERROR: unknown process\n");
+    return 0;
+  }
   
-  gSystem->Load("libEVGEN");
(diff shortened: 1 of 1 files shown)


miweberSMI:
Hi Evgeny,

I tested this new macro and it runs for me, when I am using the proper configuration. I have the following comments:
- libraries loaded only in the middle of the macro, is this wanted? Or was it always done like this?
- what happens with DIME, if a wrong process config is given?
- what is with the other processes, mentioned in the end: // Process {"pipm", "pi0", "kpkm", "ks", "rho"}

ekryshen:
Hi Michael,

* well, this is the first time we use DIME generator in the official production. I can move load libraries up to be consistent with other generators.
* I guess it will crash at the level of DIME initialization but there will be an error message printf("ERROR: uknown process\n") just before the crash. I can insert return 0 statement after that printf such that AliGenerator pointer will be NULL. Do you prefer this solution?
* "pi0", "ks", "rho" correspond to pi0-pi0, ks-ks and rho-rho production. They are supported by AliGenDime but not included in this GeneratorCustom. We don't need them for the moment (we are interested in two body charged final states). Moreover it is known that rho-rho doesn't actually work properly due to some nasty singularities in the distribution function. I think we can generalize this generator and insert it as part of GeneratorConfig.C. May be you prefer to do it now?

Thanks, cheers,
Evgeny

miweberSMI:
Hi Evgeny,

thanks for your quick reply, I would propose the following:

- well, this is the first time we use DIME generator in the official production. I can move load libraries up to be consistent with other generators.
--> I would do so, yes.

- I guess it will crash at the level of DIME initialization but there will be an error message printf("ERROR: uknown process\n") just before the crash. I can insert return 0 statement after that printf such that AliGenerator pointer will be NULL. Do you prefer this solution?
--> I would prefer to return NULL pointers, because only then it crashes with AliFatal. Otherwise it still runs with some default configuration?

- "pi0", "ks", "rho" correspond to pi0-pi0, ks-ks and rho-rho production. They are supported by AliGenDime but not included in this GeneratorCustom. We don't need them for the moment (we are interested in two body charged final states). Moreover it is known that rho-rho doesn't actually work properly due to some nasty singularities in the distribution function. I think we can generalize this generator and insert it as part of GeneratorConfig.C. May be you prefer to do it now?
--> OK, I do not think we need to do it now, one can update it later IMO.

Thanks and cheers,
Michael

ekryshen:
Hi Michael,

I didn't check what happens in DIME if the process is wrongly set. Clearly returning 0 pointer is safer. I fixed it and moved load libraries up.

Thanks, cheers,
Evgeny

PR #42, opened 2017-08-03
Added custom generator for PYTHIA + injected J/psi signals; updated t…

iarsene:
…he Generator_Jpsiee() function in MC/GeneratorConfig.C

### DIFF
--- a/MC/CustomGenerators/PWGDQ/Pythia6_Perugia2011_Jpsi2ee13TeV.C
+++ b/MC/CustomGenerators/PWGDQ/Pythia6_Perugia2011_Jpsi2ee13TeV.C
@@ -1,0 +1,20 @@
+AliGenerator *
+GeneratorCustom()
+{
+  AliGenCocktail *ctl   = GeneratorCocktail("Pythia6_Perugia2011_Jpsiee001");
+  AliGenerator   *pyt   = GeneratorPythia6(kPythia6Tune_Perugia2011);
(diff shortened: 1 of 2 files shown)


miweberSMI:
Hi @iarsene,

I have one question to the changes in the GeneratorConfig.C. What is the motivation to change the ranges and how are these found? Do you forsee future changes as well? Then it might be better to make this custom?

Thanks and cheers,
Michael

iarsene:
Hi @miweberSMI ,
   The modification of the "natural" jpsi pt range from (0,6) -> (0,1000) is not expected to change in the future. This is basically a bug fix since if one does not use a flat spectrum component, then the jpsi pt distribution ends abruptly at 6 GeV/c. We had this issue in a previous simulation unfortunately. 
     The other range which changed (extension to 50 GeV/c) is to accomodate the fact that with the EMCAL and TRD triggers we can now measure up to significantly higher pt (latest EMCAL analysis went up to 30 GeV/c). This is also not likely to change anymore, but due to the various hardcoded parameters, I would agree to make this function custom.
  Cheers, Ionut

miweberSMI:
Hi Ionut,
OK, then let's merge it for now to start the IS2017 production. But if in future we need further changes, we should make this a custom option.
Cheers,
Michael

iarsene:
Hi Michael,
Ok, I agree. Thanks a lot.
 Ionut

PR #43, opened 2017-08-05
configuration files for HF jet production in pPb with EPOS

ebruna:


miweberSMI:
Dear Elena,

I checked the configuration, but was not able to run it. There are at least two thing:
- " else printf("-- no HF process set = %s -- \n"); -> " else printf("-- no HF process set = %s -- \n",hfproc.Data());
- "hfproc" will never be filled with a string. Probably one should check "opt" to decide on the running mode.

Cheers,
Michael

miweberSMI:
I am merging this request now and will add my changes later...

PR #44, opened 2017-08-07
Addition of one more mat. bud. testing file

ddobrigk:
As discussed here: 
https://alice.its.cern.ch/jira/browse/ALIROOT-7407


miweberSMI:
Hi David,

I tested your custom generator and it works as long as the correct input is provided. One thing that I would propose, is to give a default argument for opt, i.e. (TString opt=""), in case nothing is provided. What do you think?

Cheers
Michael

miweberSMI:
Hi David,

if this is your default option, then yes, please!

Cheers,
Michael

ddobrigk:
Hi Michael, 

Done now! Thanks! 

Cheers,
David

PR #46, opened 2017-08-10
Added option for beam background simulation.

ecalvovi:
This add kDetectorBeamBkg option to remove a virtual air volume in A-Side via 
((AliPIPEv3*)PIPE)->SetBeamBackgroundSimulation();
Except that is the same as kDetectorDefault. 
Useful for Beam-Background simulations and for AD.

### DIFF
--- a/MC/DetectorConfig.C
+++ b/MC/DetectorConfig.C
@@ -13,6 +13,7 @@ enum EDetector_t {
   kDetectorMuon,
   kDetectorNoZDC,
   kDetectorCustom,
+  kDetectorBeamBkg,
   kNDetectors
 };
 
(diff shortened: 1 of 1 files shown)


pzhristov:
+1

preghenella:
@ecalvovi, 
I commented too quickly. My suggested update obviously does not work.

preghenella:
added @miweberSMI as reviewer.
The change is safe for normal AliDPG/MC operations, so it could be accepted.
I leave to @miweberSMI to proceed with this PR.

ecalvovi:
Indeed. I'm closing the request. 

Cheers,
Ernesto

PR #48, opened 2017-09-05
add pT selection in single simulation for PHOS

dsekihat:
pT selections low/mid/high are added in order to generate particle in different pT region.
This will reduce CPU time from EM showers point of view in simulation.


miweberSMI:
Hi @dsekihat,

the changes look OK to me. I will test them and then approve this request of everything is fine.

Cheers,
Michael

PR #53, opened 2017-09-20
replaced Pythia8 with Pythia6 for jetjet, removed "return NULL" in L23

ebruna:


miweberSMI:
Hi Elena,

I tested the new macro and it is running and producing outputs. So I will merge it to the master

Cheers,
Michael

PR #57, opened 2017-10-05
[ALIROOT-7487] add custom generators for dedicated Lc productions

jezwilkinson:
Adds custom generator macros for dedicated Lc productions to allow forcing of decay mode (depends on AliRoot pull number 410: https://github.com/alisw/AliRoot/pull/410)


miweberSMI:
Hi @jezwilkinson,

sorry for coming back so late to your pull request. I tested the two files, but found a few issues that need to be corrected (see attached diff file).
There is one question for the HIJING macro: you need THijing, which is initialized only later in the general GeneratorConfig.C file. This means the macro is crashing. So, I proposed another init in your custom macro, but I am not sure, if this is then overwritten later on or not. At least it was not crashing for me. 

Thanks and cheers,
Michael

jezwilkinson:
Hi @miweberSMI,
thanks for the edits, I will implement these. I must admit that I don't know about the initialisation of the Hijing object, but looking at the classes I think that a later call to Init() may well overwrite the first as the AliGenHijing::Init function calls SetMC(new THijing(...)) and replaces fHijing with a copy of that new object. Is there some functionality similar to the __R_ADDTASK__ customisation used on the lego trains that can be applied to touch the Hijing object after the existing call to Init()? Otherwise I'm not sure if it's possible to ensure these options are kept.
Thanks,
--Jeremy

miweberSMI:
Hi @jezwilkinson,

I am not aware of such a functionality in the DPG framework, we steer all configuration before intializing the generators. If you have a good idea, how to do this, I am happy to hear about it.
For now, I would propose to test it in that way and see if the configuration is kept. Do you have means to test this?

Thanks and cheers,
Michael

arossi81:
Hi Micheal ,
this message is just to confirm that, a part this pending check about whether the hijing configuration is kept after the Init() method, for me the macros by Jeremy look ok. 
@jezwilkinson , maybe just add a comment  about the line "if (randHF>0.176797){", so that anyone, also in the future can understand it. Indeed, related to this, I wonder whether it would not be better to call the first macro "Hjing_pPb_Lc.C".

Cheers,
Andrea

miweberSMI:
Hi @jezwilkinson,

thanks for the update. I tested your new update and it runs fine. If the other experts agree that the commenting out of HIJING is OK, I would just also comment out line 24 and 26 (not needed anymore).

Cheers,
Michael

PR #60, opened 2017-10-13
New custom macro for PWG-LF production with injected resonances

miweberSMI:
- ALIROOT-7529
- provided by ajay.kumar.dash@cern.ch

### DIFF
--- a/MC/CustomGenerators/PWGLF/Hijing_Rsn003.C
+++ b/MC/CustomGenerators/PWGLF/Hijing_Rsn003.C
@@ -1,7 +1,13 @@
 AliGenerator *GeneratorCustom(TString opt = "")
 {
+  //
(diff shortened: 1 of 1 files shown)


fbellini:
Hi Michael, hi Ajay, 
   I have checked the configuration and it seems to me that all is fine. 
Just as a (very minor) details, I noticed that in "ctl->AddGenerator(inj1, "Injector (Rsn003)", 1.);" and following, the string used as second argument is the same for all three particles. I think it's not a big deal, but it may be useful to have a different generator name for each species (for the logs).
That's all from my side. This can move forward and be tested.

Thanks a lot,
   Francesca

miweberSMI:
Hi @fbellini, all,

after the implementation asked by the conveners and another successful local test of the macro. I merge this into the master branch.

Cheers,
Michael

PR #63, opened 2017-10-20
update Muon configuration

pillot:
Hi Francesco,

As we discussed, I decoupled the MuonOnly and MuonCalo config.

For the MuonCalo, I explicitly disabled the tasks that I think cannot run with this reconstruction configuration (MUON ITS VZERO T0 AD), while letting the others as defined by default, assuming that if the default change it is also fine for MuonCalo.

I ran 2 tests locally to check:
i) that the config MuonOnly is unchanged
ii) that the config MuonCalo is OK, as far as I can tell... i.e. OK for muon tasks and it does not crash

Since I am not familiar with all the tasks in the train, can you please cross-check the MuonCalo config before merging?

Thanks and cheers,
Philippe

### DIFF
--- a/QA/QAtrainsim.C
+++ b/QA/QAtrainsim.C
@@ -80,7 +80,7 @@ Bool_t doAD           = 1;    //decetrot AD
 Bool_t doEvTrk        = 1;    // analysis task uses the CF framework 
 
 Bool_t isMuonOnly     = kFALSE; // setting this to kTRUE will disable some not needed tasks for a muon-only MC
-Bool_t isMuonCalo     = kFALSE; // setting this to kTRUE will disable some not needed tasks for a muon-calo MC (V0,MUON,SPD)
+Bool_t isMuonCalo     = kFALSE; // setting this to kTRUE will disable some not needed tasks for a muon-calo MC (MUON ITS VZERO T0 AD)
 
 Int_t debug_level     = 1;    // Debugging
 Int_t run_number      = 0;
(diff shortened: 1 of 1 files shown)


pillot:
Hi Francesco,

I am not quite sure about the use cases for the MuonCalo configuration. I think so far only UPC people used it and in my mind it is meant for e.g. Pythia, STARLIGHT or other full MC (contrary to MuonOnly where we generate only muon signal, e.g. jpsi, for AccEff of detector performance studies). 

So my understanding is that this settings is supposed to be close to the muon_calo data productions, which is why I started from the default settings, removing only what cannot work like TPC stuff etc... In particular, I think the physics selection can be activated here. Only in the MuonOnly case the PS cannot work, because only muon is reconstructed. Idem for the centrality. Since the full ITS, T0 and AD are reconstructed, I was thinking to also let the QA activated for those detectors.

But if you prefer to disable everything like in MuonOnly and only activate a few tasks, it's fine for me too.

About the SelectCollisionCandidates for muon performance and efficiency, only for MuonOnly it cannot work. For the rest, I assumed the physics selection would select the proper events to analyse, which is why I kept this selection like for other tasks. But if we disable the PS, for sure we have to disable this selection as well.

Cheers,
Philippe

fprino:
Hi Philippe,

OK. Thanks for the explanations!
I'll merge your changes and then we can test the new macro on one of the output of the recent simulation that was done with the MuonCalo configuration for the multiplicity analysis.

Cheers, Francesco

PR #69, opened 2017-11-07
exit with 0 only on successful completion

shahor02:


fprino:
Thanks you Ruben for this fix.
I have ported your commit to v5-09-XX and v5-08-XX branched, from which we tag the versions for productions.

Cheers, Francesco

shahor02:
Thanks!

PR #71, opened 2017-11-16
Added custom generator setup for Hijing and injected jpsi for Xe-Xe

iarsene:


miweberSMI:
Hi @iarsene,

I am checking your new custom generator macros and have two comments:
- In the HIJING+STARLIGHT macro: "cocktail" or "ctl" (at the moment both is used and crashing)
- In general, I would try to name the macros more consistently: either with numbers or without. But probably it is already too late for that and we coudl fix that in some "cleaning mission" later on.

Cheers,
Michael

miweberSMI:
Hi again,

furthermore I see that you exchange one line:
 +  if(!processConfig.Contains("TwoGamma")) genCocktail->AddGenerator(genEvtGen,"EvtGen",1.); 
only adding if the process is not TwoGamma. Is this wanted? Was this done wrongly before?

Thanks for calrification,
Michael

iarsene:
Hi Michael,
Regarding the "cocktail" vs "ctl", sorry if i managed to screw this up. In my version i have "cocktail"  everywhere. Maybe i managed to commit the uncorrected version.
  On the line exchange, that is intentional. The EvtGen is used for the proper decay of charmonia (including the QED radiation). For the gamma+gamma process one does not need EvtGen.
  Cheers, Ionut

miweberSMI:
Hi Ionut,

I tested both new generator configurations and they seem to run fine.
I also cross-checked with conveners that the macros are OK in the current form.

Thanks and cheers,
Michael

PR #74, opened 2017-11-23
Changed the order of adding the generators to the cocktail.

iarsene:


miweberSMI:
Hi Ionut,

I tested this change and it is running. I will merge.

Thanks and Cheers,
Michael

PR #77, opened 2017-11-26
Resonance + Nuclei injection scheme

mpuccio:
Hi @miweberSMI ,

this configuration will be needed for an upcoming production request by PWGLF.
We still need to put some new states that are not present in the current AliRoot release, thus expect an update on this!

Thanks and cheers!
Max

### DIFF
--- a/MC/CustomGenerators/PWGLF/Pythia8_Monash2013_RsnNcl001.C
+++ b/MC/CustomGenerators/PWGLF/Pythia8_Monash2013_RsnNcl001.C
@@ -1,0 +1,41 @@
+struct particle_inj {
+  int pdg;
+  double maxpt;
+  double maxy;
+};
+
+AliGenerator *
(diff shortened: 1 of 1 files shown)


miweberSMI:
Thanks @mpuccio 
I will test it and approve it in case the test is succesful (but only then).

Cheers,
Michael

miweberSMI:
Hi @mpuccio 

I tested it and it is running fine.
Thanks for preparing this nice generator macro!

Cheers,
Michael

miweberSMI:
Hi again @mpuccio,

I ported the commit also to the production branches for v5-08-XX and v5-09-XX, so as soon as we have a new tag, you should be able to test it on the GRID. 

Have a good night,
Michael

mpuccio:
Thanks a lot!

miweberSMI:
You are welcome!

PR #80, opened 2017-12-06
Background for MC-2-MC embedding can be taken from .xml collection

preghenella:


miweberSMI:
Hi @preghenella,

I looked at your commit and agree with it (as far as I could understand what is done here). I will merge it. To which production branches we need to port it?

Thanks and cheers,
Michael

preghenella:
this is a generic commit, it can go in all production branches.

miweberSMI:
done! 

Thanks again,
Michael

PR #83, opened 2017-12-13
New PWG-LF custom generator

miweberSMI:
- DPMJET + injected resonances for p-Pb 8 TeV
- Author: rama.chandra.baral@cern.ch

### DIFF
--- a/MC/CustomGenerators/PWGLF/DPMJET_RsnInject_pPb8TeV.C
+++ b/MC/CustomGenerators/PWGLF/DPMJET_RsnInject_pPb8TeV.C
@@ -1,0 +1,76 @@
+
+AliGenerator *GeneratorCustom(TString opt = "")
+{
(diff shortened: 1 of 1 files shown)


fbellini:
Hello Michael, Rama,
thanks for preparing the macro.
The generator is ok and it is indeed different wrt the generator for the p-Pb 5 TeV, as here the injection is not done flat in pT contrary to there. 

About the naming, I would stick to the convention "<generator>_Rsn_<system><index>.C", e.g. "Dpmjet_Rsn_pPb002.C". I would not specify the energy in the macro name, because there is nothing specific of a given energy in the configuration. In this way, we can have several generators ready to be used with different energies. The system is instead informative, because the rapidity range is set accordingly.

We then give green light to port this macro in AliDPG. Can you please take care of changing the name though?

Thank you and cheers,
Alexander and Francesca

miweberSMI:
Hi Francesca,

thanks for your reply and confirmation, the commit is merged and in another commit I changed already the names accordingly.

Thanks and cheers,
Michael

PR #84, opened 2017-12-15
DPMJET f0 and f2 injected pPb5TeV

ekryshen:
By request from Viktor Riabov

### DIFF
--- a/MC/CustomGenerators/PWGLF/DPMJET_RsnInject_pPb5TeV.C
+++ b/MC/CustomGenerators/PWGLF/DPMJET_RsnInject_pPb5TeV.C
@@ -1,0 +1,35 @@
+AliGenerator *GeneratorCustom(TString opt = "")
+{
+  // Custom generator to inject f0 and f2
+  // Options:
+  // pPb:  flat pT distibution in a range 0-15 GeV/c; 4 particles per event;
+  // rapidity range: -0.6 < y < 0.1 : Anchored to period LHC16q
+
+  Int_t ninj = 4; // 4 injected particles per species
+
+  // Set pT range (pT_L, pT_H)
+  Double_t  pT_L = 0.0;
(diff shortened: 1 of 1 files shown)


vriabov:
The request is related to JIRA ticket:
https://alice.its.cern.ch/jira/browse/ALIROOT-7633?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&focusedCommentId=208069#comment-208069

miweberSMI:
Hi Viktor,

thanks for the pull request. I tested it and it is working fine. Before merging and porting to the production branches, I would like the conveners to comment. Especially o the macro naming, since we have a similar one (with quite different structure and resonances) here:
https://github.com/alisw/AliDPG/blob/master/MC/CustomGenerators/PWGLF/DPMJET_RsnInject_pPb8TeV.C

Cheers,
Michael

fbellini:
Hello Michael, Victor,
    thanks for preparing the macro. 
The generator is ok and it is indeed different wrt the generator for the p-Pb 8 TeV, as the injection is done flat in pT contrary to there. It is done similarly to what done in past productions with hijing and pythia. 

About the naming, I would stick to the convention <generator>_<rsn>_<system><number>, e.g. "Dpmjet_Rsn_pPb001.C". I would not specify the energy in the macro name, because there is nothing specific of a given energy in the configuration. In this way, we can have several generators ready to be used with different energies. The system is instead informative, because the rapidity range is set accordingly.

We then give green light to port this macro in AliDPG. Can you please take care of changing the name though?

Thank you and cheers,
   Alexander and Francesca

miweberSMI:
Hi Francesca,

thanks for the comment and approval. I merged it to the AliDPG master, and will change the name of this (and the other macro accordingly).

Cheers,
Michael

PR #85, opened 2017-12-19
 AliEn -> local copy of the background files for embedding

preghenella:


miweberSMI:
Hi @preghenella,

looks good to me, I will wait for @catalinristea to comment as well and then merge it. Do we need this in both production branches (08 and 09)?

Thanks and cheers,
Michael

preghenella:
the PR is valid for both branches

catalinristea:
Hi Roberto, Michael

Everything looks good from what I see. I also like the idea of the separate zip archive for the SDigits, to be present in JDL and easily identifiable. 

If we would have had the JDL place holder, we could have had specified directly into the wn.xml the archives, and get them downloaded before even starting to run, dropping easy ERROR_IB in case there are problems to download the large archives - this way we'll probably get ERROR_V, which is also fine. I'm not sure if we need anything in the dpgsim.sh for the case the zip is corrupted or not downloaded correctly - but such things we could fix after the tests.

Thanks!

Yours,
Catalin.

miweberSMI:
Hi,

I just merged the PR and will port it to both production branches.

Thanks,
Michael

PR #87, opened 2018-01-09
Add files via upload

adubla:


adubla:
new Hijing_Xe_HF.C for HF production in XeXe

miweberSMI:
Hi @adubla,

thanks for uploading your macro. I tested it and it works fine as far as I can tell (which means the simulation is running).
One thing I am wondering, is that you removed the random inclusion of HIJING events. I seem to remember that this was always needed for HF productions, but I might be wrong. Perhaps the conveners can comment? I would also need their OK for the macro naming in any case.

Thanks and cheers,
Michael

adubla:
Hi Michael,

i removed the random inclusion of HIJING because this was just there for the case of pPb. For the PbPb pr in this case XeXe is not needed.
The convener can comment more in case necessary.

Cheers,
Andrea

arossi81:
Thanks Deepa and @adubla for preparing the file and making the pull request. @miweberSMI : the Config is fine for me and Alessandro. 
Cheers,
Andrea

PR #92, opened 2018-01-19
D2H pp production with high mult cut

yowatana:
CustomGenerator for the analysis of high-multiplicity pp events from PYTHIA.


miweberSMI:
Hi @yowatana,

I tested the macro and it runs fine for me. Before merging it, I would like the conveners to confimr the naming of the macro here. Could you ask them to do so?

Thanks and cheers,
Michael

arossi81:
Hi @miweberSMI,
yes, the name is ok for us. Yosuke had already iterated on this with us.
Cheers,
Andrea

PR #93, opened 2018-01-22
Adding Therminator2 generator

maciekniewielki:
This pull request adds the Therminator2 generator script, along with the configuration script. More information about the configuration can be found [here](https://github.com/alisw/AliPhysics/pull/3867).

I also changed the GeneratorConfig.C to include Therminator2 in the generator list.


miweberSMI:
Dear @maciekniewielki,

thanks a lot for uploading this Pull Request.
After discussing with other DPG members and Latchezar, we have a concern about the usage of Python to set the parameters. If that Python is needed for that, we have to start carrying it in the ALICE CVMFS... Some python is available on the nodes, but version not guaranteed. Perhaps one can do the parameter setting in  a different way?

Cheers,
Michael

pzhristov:
I think we have never requested python on the worker nodes. Please use bash as standard scripting language.

Cheers,
Peter

maciekniewielki:
I see. If that's the case, then I will port the python script to bash.

I also wanted to ask about the available tools. Can the script use standard POSIX command line tools, such as sed?

Cheers,
Maciej

pzhristov:
Hi @maciekniewielki,

Sed and awk should be available on each worker node.

Cheers,
Peter

maciekniewielki:
Hi @pzhristov, @miweberSMI,

I got rid of the python script. Its functions have been ported to the gen_therm2.sh file. 

Cheers,
Maciej

pzhristov:
+1

miweberSMI:
Hi Maciej,

thanks for the update. It looks good to me and I am testing it on lxplus right now, in case it runs fine I will merge the PR. One question that I still have is, how would you set the parameters that are mentioned in the script (if different from default)?

Thanks and cheers,
Michael

maciekniewielki:
Hi Michael,

The non-default parameters should be set by changing the corresponding THERM2_PARAMS_<parameter_name> environment variable. It must be done before running the generator.

For example, the default freezeout model is Lhyquid2DBI (as described in [events.ini](https://github.com/alisw/therminator/blob/alice/v2.0.3/events.ini) configuration file). If we want to change it to the Lhyquid3D, the environment variable THERM2_PARAMS_FreezeOutModel should be set to "Lhyquid3D" (the FreezeOutModel parameter name is used in the events.ini file).

Therminator2 uses separate configuration files for different models, so the configuration of the Lhyquid3D model is in the [fomodel/lhyquid3d.ini](https://github.com/alisw/therminator/blob/alice/v2.0.3/fomodel/lhyquid3d.ini) file. Suppose we chose this model and now want to change the rapidity range from the default 12.0 to 14.0. We must set the THERM2_PARAMS_RapPRange variable to "14.0".

This process of parameter setting can be done using any languages that support assignment to environment variables. It can be done using "export" in bash or using "gSystem->Setenv" in ROOT. Setting these two parameters in ROOT would look like this:
`gSystem->Setenv("THERM2_PARAMS_FreezeOutModel", "Lhyquid3D");`
`gSystem->Setenv("THERM2_PARAMS_RapPRange", "14.0");`

There is also a special parameter THERM2_PARAMS_XML_PATH, which doesn't appear in the Therminator2 configuration. It is used to download (via alien_cp) and set a custom FreezeFile for the hydro models. I've added a sample usage in the GeneratorConfig.C file.

Let me know if you have any more questions.

Cheers,
Maciej

miweberSMI:
Hi Maciej,

OK, thanks for further explanations. Have you already thought about how this could be done in the "AliPDG framework"? Probably the easiest would be different "custom macros" that set exactly these non-default values if needed.
The tests on lxplus finished and were successful (at least not crashing), but I realized that Therminator alone already needs ~1.5h for one event. Is this expected?

Thanks and cheers,
Michael

maciekniewielki:
Hi Michael,

I thought that the configuration part is done in GeneratorConfig.C, specifically in the functions that return AliGenerator pointers. Apart from parameters defined in "Monte Carlo generation control" [here](https://twiki.cern.ch/twiki/bin/viewauth/ALICE/AliDPGMonteCarloTutorial) I've seen some hardcoded values in these functions, so I assumed that this macro is changed each time a different configuration is needed. This would mean that the Therminator2 parameters could be changed by just changing a few lines in the GeneratorTherminator2 function. If it is not the case, can you elaborate more on where (and how) the configuration of these hardcoded values is taking place?

As for the long generation time, it is normal for Therminator2. At the start of the event generation it first has to calculate max integrand value and the primordial multiplicity, which takes around 1-1,5h. Only then Therminator2 begins generating events. One event generation takes a couple of seconds for the default configuration. Because of that, generating only a few events at a time is not efficient as most of the time is taken by these initial calculations. This time can be lowered for testing purposes by setting the IntegrateSamples parameter (it is defined in the events.ini file).

Thanks,
Maciej

miweberSMI:
Dear Maciej,

OK, thanks for clarification. We can discuss then other parameter settinsg when its needed (perhaps easier via Skype/Vidyo/Telephone).
For the long initialization time, is there a possibility to speed up? We have to see in tests if this affects the performance.
In any case, I will merge now, so that we have this in a new tag and can start the full production testing.

Thanks and cheers,
Michael

maciekniewielki:
Dear Michael,

In order to lower initialization time, one should lower the IntegrateSamples parameter. It should scale linearly with initialization time. The default value in events.ini is 5000000, so lowering it to 1000 or 100 should make the initialization time negligible. It can be set like any other parameter, for example:
`gSystem->Setenv("THERM2_PARAMS_IntegrateSamples", "1000");`

Please bear in mind that lowering it this much will negatively affect the quality of data, so it should only be done for testing purposes.

Cheers,
Maciej

PR #94, opened 2018-01-24
Setup macro for PWGDQ jpsi2ee production pp 5TeV (2017 data)

mkkohler:
Generator Pythia6, perugia 2011
Natural pT shape for 5 TeV pp data
Runlist LHC17p,q


miweberSMI:
Dear @mkkohler,

I tested the macro and it runs for me. So I would merge it, but first I would like th econveners to comment on it here (all is as requested and the naming is fine).

Thanks and cheers,
Michael

miweberSMI:
Dear all,

got the confirmation from the PWG-DQ conveners, that everything is fine with this macro. On the naming, we can still change it at a later stage, perhaps needs some more general discussion.

Cheers,
Michael

PR #96, opened 2018-01-27
Config for HF jets production in pPn with HIJING

ebruna:


miweberSMI:
Dear @ebruna, all,

I tested the macro locally and didn't observe any problems. Assuming that the content and macro naming was discussed with the conveners, I merge the macro and port it to the production branches.

Thanks and cheers,
Michael

PR #98, opened 2018-02-02
 Custom generator for jpsi injected MC for p-Pb at 8TeV updated

iarsene:


miweberSMI:
Hi @iarsene, all,

I tested the updated macro and it runs fine (on lxplus without crashing). I also cross-checked with conveners that the changes are OK like that. For that reason I am merging the PR and port it to the production branches.

Cheers,
Michael

PR #102, opened 2018-02-12
 First version of MC-to-MC embedding version for injected signals

miweberSMI:


miweberSMI:
After green light from the PWG responsibles for these macros, I am merging this PR.

PR #103, opened 2018-02-12
Fixes for ROOT 6, disable some stuff which I could not make run easily

davidrohr:
This is only for testing and discussion right now, please do not merge yet

### DIFF
--- a/MC/Config.C
+++ b/MC/Config.C
@@ -8,6 +8,10 @@
 /*****************************************************************/
 /*****************************************************************/
 
+#if !defined(__CLING__) || defined(__ROOTCLING__)
+#include "TGeant3TGeo.h"
+#endif
+
 // global variables
 
 static Int_t   runNumber       = 0;         // run number
@@ -16,7 +20,7 @@ static Int_t   magnetConfig    = 0;         // magnetic field
 static Int_t   detectorConfig  = 0;         // detector
 static Int_t   generatorConfig = 0;         // MC generator
 static Float_t energyConfig    = 0.;        // CMS energy
-static Float_t triggerConfig   = 0.;        // trigger
+static Int_t triggerConfig   = 0.;        // trigger
 static Int_t   pdgConfig       = 0;         // PDG value 
 static Float_t bminConfig      = 0.;        // impact parameter min
 static Float_t bmaxConfig      = 20.;       // impact parameter max
@@ -39,6 +43,13 @@ static Int_t   quenchingConfig = 0;         // quenching
 static Float_t qhatConfig      = 1.7;       // q-hat
 static Bool_t  isGeant4        = kFALSE;    // geant4 flag
 
+#include "DetectorConfig.C"
+#include "GeneratorConfig.C"
+
+void ProcessEnvironment();
+void CreateGAlice();
+void GeneratorOptions();
+
 /*****************************************************************/
 
 void
@@ -46,8 +57,6 @@ Config()
 {
 
   /* initialise */
-  gROOT->LoadMacro("$ALIDPG_ROOT/MC/DetectorConfig.C");
-  gROOT->LoadMacro("$ALIDPG_ROOT/MC/GeneratorConfig.C");
   ProcessEnvironment();
 
   /* verbose */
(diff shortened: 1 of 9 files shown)


chiarazampolli:
Ciao @davidrohr ,

I would like to merge your PR, is it ok if I have a quick look and then merge?

Chiara

davidrohr:
Ciao @chiarazampolli ,
I am not sure if we do not want to fix starlight and dpmjet before, or perhaps add a check for the root version instead of just commenting it out.
But sure, from my side you can merge it.

chiarazampolli:
Ciao David,

Thanks! I will have a look tomorrow, but I would not comment out anything, just merge what you did, start with what we have, and wait for STARLIGHT and DPMJET to have them also working under ROOT6. @pzhristov said he would like to check those. 

Cheers,

Chiara

davidrohr:
Ciao Chiara, just to be clear: In my PR starlight and dpmjet are commented out. If you merge this, they will not work any more.

One should change this to a ROOT version check instead of unconditionally disabling them.
Or you merge and fix it later, but then in the meantime starlight and dpmjet won't work.

chiarazampolli:
I understood as you write, that these generators are commented out. Since I am not the author of the MC part, and I never really looked at it, and I have not yet checked your PR, I don't know if the code there is one single macro where you have all generators, or if STARLIGHT and DPMJET are independent. This is what I wanted to check tomorrow. From your comment I understand that everything is together, so you need to take action (either commenting, or putting the "if" on the ROOT version) to have things work (with or without STARLIGHT/DPMJET). If this is the case, then I suggest the "if" approach, like:

#if ROOT_VERSION_CODE >= ROOT_VERSION(6,0,0)

(which was also the suggestion by Ruben to overcome an issue somewhere else). I believe this is also what you would do. For me you can go ahead with this change too, or I will add it after merging the PR.

But as I said, I planned to review your PR tomorrow, so sorry in case I misunderstood.

davidrohr:
Ciao @chiarazampolli ,

To save you some work, I just rebased my patch onto the newest master, and added #ifdefs for the ROOT version instead of just disabling dpmjet and Starlight.

I verified that this works for me now with ROOT6.
I did not test with ROOT 5.

Cheers
David

chiarazampolli:
Ciao David,

Many thanks! I will review the PR tomorrow and merge in case i have no comments.

Good night,

Chiara

chiarazampolli:
Ciao David,

I have a few comments/questions, see below. 
1) Do you need the lines:

`+#include "DetectorConfig.C"`
`+#include "GeneratorConfig.C"`

in  MC/Config.C?

2) I saw in a couple of places these lines:

`#if` !defined(__CLING__) || defined(__ROOTCLING__)`
`+#include "AliGenPythia.h"`
`#endif`

I guess it is needed... 

3) Did you try to execute:

`gROOT->ProcessLine("gg_tmp_gen = GeneratorCustom(gg_tmp_str);");`

To me it looks absolutely fine, but I am asking to be sure, as in such cases I was using a different approach. Yours is much simpler, so I would be happy to use it next time. 

4) Did you try this change:

`-    mcHandler->SetPreReadMode(1);`
`+    mcHandler->SetPreReadMode((AliMCEventHandler::PreReadMode_t) 1);`

Maybe we can have a look in person, it should be faster than through github :) 

Thanks!

Chiara
both with ROOT6 and ROOT5? I had issues with enums, and had to implement the following:

` #if ROOT_VERSION_CODE >= ROOT_VERSION(6,0,0)`
`    AliAnalysisTaskVertexESD* taskvertexesd =  AddTaskVertexESD(kFALSE, static_cast<AliVEvent::EOfflineTriggerTypes>(kTriggerMask)); //ROOT6`
`    #else`
`    AliAnalysisTaskVertexESD* taskvertexesd =  AddTaskVertexESD(kFALSE, kTriggerMask); // ROOT5`
`    #endif`

davidrohr:
Ciao Ciara, I won't have time today but we can sit together on Moday.

The cling #ifdefs allow for the usage of the macro in compiled and in interpreted mode. For AliDPG it is not strictly necessary but it simplified the debugging.

The gROOT ProcessLine approach worjs and is pretty simple, but it has the disadvantage that it needs these global objets.
Kind Regards
David Rohr

(Sent from my mobile, excuse the typos.)

chiarazampolli:
Hello,

After we discussed yesterday, I am merging this PR.

Cheers,

Chiara

PR #104, opened 2018-02-15
Split resonance and nuclex MC productions

mpuccio:


miweberSMI:
hi @mpuccio,

I tested the macro and it runs fine in the master. I am merging this PR and will port it to the production branches. One question that I had, is where the resonances will go, but perhaps I misunderstood something here...

Cheers,
Michael

PR #109, opened 2018-03-11
Small modifications for TOF in data taking

miweberSMI:
- correction: removing "$" for setting CONFIG_TRIGGER in dpgsim.sh
- setting from outside should have priority: changing "else" into "elif" statement
- recognize Custom config file from ALIDPG directory: adding additional if statements in "Config.C"

### DIFF
--- a/MC/Config.C
+++ b/MC/Config.C
@@ -215,6 +215,13 @@ ProcessEnvironment()
 	break;
       }
     if (!valid) {
+      TString triggerConfigString(gSystem->Getenv("CONFIG_TRIGGER"));
+      if(triggerConfigString.Contains("Custom.cfg")){
+	triggerConfig = kTriggerCustom;
+	valid = kTRUE;
+      }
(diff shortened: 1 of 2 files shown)


chiarazampolli:
Ciao Michael,

Thanks for the fixes :) When I tested, I didn't realize the bug in the extra "$" as I made just an external script, and I was then not using the variable at all. :) 

Cheers,

Chiara

miweberSMI:
Hi @chiarazampolli, all

to your question: to be honest, I do not know.
There was still the problem that the Trigger.root was not filled. Now I used a simpler solution, just copying the Custom.cfg file into the directory. Now everything seems to work. If you agree I would merge it to the master?

Cheers,
Michael

chiarazampolli:
Hi Michael,

I agree, let's do as you suggest for now, but I think we need to understand what is not working...

Thanks,

Chiara

PR #114, opened 2018-03-16
Fix string size in struct for the resonance generator

mpuccio:
Hi @miweberSMI @catalinristea ,

this fixes the issue discovered on the grid for the resonance generator.
In my local tests it is working, but it was the case also for the previous one... This time I checked also on lxplus and it should be OK. Please have a look!

Cheers,
Max

### DIFF
--- a/MC/CustomGenerators/PWGLF/Pythia8_Monash2013_Rsn002.C
+++ b/MC/CustomGenerators/PWGLF/Pythia8_Monash2013_Rsn002.C
@@ -1,6 +1,6 @@
 struct particle_inj {
   int n;
-  TString name;
+  char name[32];
   int pdg;
   double maxpt;
   double maxy;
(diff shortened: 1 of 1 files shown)


fprino:
Ciao Max,

I saw that you used Form(), which may show erratic behaviour in rare occasions (this is what I learned from Dario...) I was wondering if you could do the change using a safer implementation:
Dario suggested in a different occasion these 2:

  TString s;
  s.Form("%p", asd);
or:
  TString s = TString::Format(...)


Ciao, Francesco

mpuccio:
Ciao Francesco,

OK I changed the code according to your comment!
If the nuclei production are affected by the same issue I can fix also the other generator.

Cheers,
Max

fprino:
Thanks Maximiliano.
I'll merge now your pull request to the master, so that Catalin can use the modified version to run a quick test on grid.

PR #116, opened 2018-03-18
Add generator for Herwig7

mfasDa:
Events are generated with the Herwig7
executable in an external process and
read in AlRoot with AliGenReaderHepMC.
Config files for Herwig are handled
as template files replacing instances
of CONFIG* with the corresponding
environment variables set by dpgsim.sh.
Translation of the templates into
Matchbox files is done using JINJA.
Initially a template is provided for
Min. Bias pp, more templates eventually
to be added in the future.

### DIFF
--- a/MC/CustomGenerators/PWGJE/LHC-MB.tmpl
+++ b/MC/CustomGenerators/PWGJE/LHC-MB.tmpl
@@ -1,0 +1,53 @@
+# -*- ThePEG-repository -*-
+
+################################################################################
+# This file contains our best tune to UE data from ATLAS at 7 TeV. More recent
+# tunes and tunes for other centre-of-mass energies as well as more usage
+# instructions can be obtained from this Herwig wiki page:
+# http://projects.hepforge.org/herwig/trac/wiki/MB_UE_tunes
+################################################################################
+
+##################################################
+# Technical parameters for this run
+##################################################
+cd /Herwig/Generators
+set LHCGenerator:NumberOfEvents {{ CONFIG_NEVENTS }} 
+set LHCGenerator:RandomNumberGenerator:Seed {{ CONFIG_SEED }}
+set LHCGenerator:PrintEvent 10
+set LHCGenerator:MaxErrors 1000000
+
+set LHCGenerator:DebugLevel 0
+set LHCGenerator:DumpPeriod -1
+set LHCGenerator:DebugEvent 0
+
+
+##################################################
+# LHC physics parameters (override defaults here) 
+##################################################
+set LHCGenerator:EventHandler:LuminosityFunction:Energy {{ CONFIG_ENERGY }}
+
+# Intrinsic pT tune extrapolated to LHC energy
+set /Herwig/Shower/Evolver:IntrinsicPtGaussian 2.2*GeV
+
+##################################################
+# Matrix Elements for hadron-hadron collisions 
+##################################################
+cd /Herwig/MatrixElements/
+insert SimpleQCD:MatrixElements[0] MEMinBias
+
+
+# MPI model settings
+set /Herwig/UnderlyingEvent/MPIHandler:IdenticalToUE 0
+cd /Herwig/Generators
+
+# Change to have no pT cuts for MinBias
+set LHCGenerator:EventHandler:Cuts /Herwig/Cuts/MinBiasCuts
+
+insert LHCGenerator:AnalysisHandlers 0 /Herwig/Analysis/HepMCFile
+set /Herwig/Analysis/HepMCFile:Format GenEvent
+set /Herwig/Analysis/HepMCFile:Units GeV_mm
+set /Herwig/Analysis/HepMCFile:PrintEvent {{ CONFIG_NEVENTS }}
+set /Herwig/Analysis/HepMCFile:Filename output.hepmc
+
+run LHC-MB LHCGenerator
+
--- a/MC/EXTRA/gen_herwig.sh
+++ b/MC/EXTRA/gen_herwig.sh
@@ -1,0 +1,19 @@
+#! /bin/bash
+set -e
+
+if [[ $# -lt 1 ]]; then
+    echo "usage: gen_eposlhc_PbPb.sh [outfilename] [nevents] [projectile_id] [projectile_energy] [target_id] [target_energy]"
+    exit -1
+fi
+
+echo "running in `pwd`, writing HepMC to $1"
+
+# prepare environment
+source /cvmfs/alice.cern.ch/etc/login.sh
+eval $(alienv printenv Herwig/v7.0.4-alice1-3)
+export LHAPDF_DATA_PATH=$LHAPDF_ROOT/share/LHAPDF
+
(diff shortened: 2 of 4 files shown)


mfasDa:
Template handling for Herwig configurations discussed here: https://indico.cern.ch/event/713769/contributions/2932699/attachments/1616161/2570651/HerwigTemplateEngine.pdf

miweberSMI:
Hi @mfasDa, 

thanks for this PR and the detailed information about input file creation. Indeed this could be useful also for other external generators. For example, we were discussing similar things when implementing the new AMPT or Therminator generators. During that discussion, the offline team strongly recommended not to use Python etc. since this would mean that additional packages need to be provided on the nodes. So my question now, was this discussed already with Offline (perhaps @chiarazampolli or @fprino know more about it than me at this moment)? 

Thanks and cheers,
Michael

mfasDa:
Hi @miweberSMI, 

The outcome of the discussion was that this will be brought to the computing board the next occasion by the PWGMM. Until this time the PR should be considered as proposed change and not yet to be merged!

Honestly I don't see the problem that offline mentions - the only module required is in any case a dependency of other modules and therefore implicitly included in the python-modules package (a package which is needed in any case for ROOT6 builds), and the usage of JINJA significantly simplifies the configuration handling.

Note that it was discussed in the PWGMM meeting to move the template files to GenMC, which should work similarly well, however for fast development I would argue to keep it at the moment in AliDPG.

I did a local test production with full Herwig7 chain via dpgsim and it was working as desired.

Cheers

Markus

rbertens:
Hi @miweberSMI ,

We have discussed the proposed changes form @mfasDa  at the previous PAG-MC meeting ( https://indico.cern.ch/event/713769/ ) , @qgp will propose them at the upcoming offline meeting. 

My best,

Redmer

mfasDa:
Several weeks have passed now (one file even got out-of-sync :( ) Any progress with this PR? Has the usage of python + jinja been discussed in the computing board meting as promised in the PWGMM meeting?

chiarazampolli:
Hello @mfasDa ,
No, there was no discussion at the CB, and since we understood that this was the prerequisite for this PR, we didn't look into it. Please, contact MM. 

Chiara

qgp:
Hi,

from the discussion we had, it turned out that jinja is already pulled in as an implicit dependency of python-modules. The only thing remaining is to make this dependency explicit by putting jinja in the list. Unfortunately, with the recent activities either I couldn't attend or the offline meeting was cancelled, so this request still is to be discussed. Anyway, as jinja is there already, this doesn't block any of the testing.

Other than that, I would argue that it makes sense to do the testing right away with aligenmc, then everybody (also standalone usage for Rivet) would benefit from these developments and we would have consistency right from the beginning.

Cheers,
   Jochen

miweberSMI:
Dear Jochen, dear Markus,

sorry for coming back so late on this...
I just tried a simple test on lxplus with this pull request, but it seems the simulation crashes. When using the AliDPG master, this doesn't happen. Could you fix this (in parallel to the discussion on the Python/Jinja usage)?

Thanks a lot and cheers,
Michael

mfasDa:
Dear @miweberSMI ,

Could you please expand a bit more what is actually crashing? An existing (i.e. PYTHIA or AMPT) generator or did you try to use Herwig7?

Concerning 1st) Should not happen because the HERWIG generator works in a separate function inside the config, completely separated from other code (so should not be called at all), and I adapted all arrays. So would be good to see a stack trace or a log indicating what is the problem (what was overlooked).

Concerning 2nd) Indeed when I tested the changes I observed a crash as well, happening in the AliGenHepMCReader. After debugging a bit it turned out that the numbers of particles in the HepMC event (incoming) and the output array were different (more particles in the HepMC event), and since the HepMC event was used to return the array size a crash happened at a later stage due to buffer overflow. I was able to run through by fixing the particle collection size to use the size of the output array and not the one from the HepMC event, but that of course indicates an issue that needs to be investigated why the Herwig event has more particles than what is imported into the stack.

Best Regards

Markus

miweberSMI:
Hi Markus,

I tried Herwig and a standard generator. Could it be due to the unresolved conflicts?

Thanks and cheers,
Michael

miweberSMI:
Hi @mfasDa,

here is the sim.log file. I apologize that I didn't spend mor etime at the moment to look into it: 
[sim.log](https://github.com/alisw/AliDPG/files/1936059/sim.log)

Cheers,
Michael

mfasDa:
Hi @miweberSMI,

Thanks for sharing the sim.log! From this it becomes clear that the reason for the crash is not related to my commits but due to the root6 fixes (something I also observed in local tests I did to verify the fixes in the EMCAL simulation and tried to address in #125). The problem comes from the interpretation of the Config.C macro (AliSimulation Line 1116), where the real hint to the problem can be found here: 

```
Error: cannot open file "TGeant3TGeo.h"  /afs/cern.ch/user/m/miweber/DPG/AliDPG//MC/Config.C:13:
*** Interpreter error recovered ***
Error: can not call private or protected function (tmpfile):1:
```

What I realized in my test is that in pretty much all macro for code protected in `#if ROOT_VERSION_CODE`  the comparison operator is swapped so that ROOT6-related code is executed under ROOT5 and vice versa. Without my fixes this code (the sim part) neither ran for me under ROOT5 nor under ROOT6, with my fixes I managed to run under both (apart from the geant4 part - that needs another look). Also I commented that I would recommend to rather use the macro `__CLING__` to distinguish ROOT5 and ROOT6 since the separation like this will work both under ROOT6 and ROOT5 in interpreted mode (tested i.e. with the handlers macro in the lego train, ROOT_VERSION_CODE seems to not be defined in ROOT5 interpreted mode).

Your problems are not related to this pull request!

Cheers

Markus

mfasDa:
Hi @chiarazampolli ,

Thanks!

For what concerns the discussion of distinguishing ROOT5 and ROOT6 this becomes now hijacking this PR for the discussion of an unrelated topic (this PR is about Herwig integration!) - can we discuss this issue in a more suited place (i.e. JIRA ticket for ROOT6 migration - there exists already a super-ticket and would just need a sub-ticket for AliDPG)?

Cheers

Markus

miweberSMI:
Hi @mfasDa, @qgp, all,

following our discussion last week, should we close the PR here and wait for the development on aligenmc?

Thanks for clarifying and cheers,
Michael

miweberSMI:
Hi @mfasDa,

 we have now (sorry for taking so long!) a running aligenmc+AliDPG, which means we could run Herwig, once it is implemnted in aligenmc (I guess this is then a question to @qgp and Antonin Mairie?).
Can we close this PR then?

Thanks and cheers,
Michael

mfasDa:
Hi @miweberSMI,

Indeed if Herwig simulations are supported via aligenmc then indeed there is no need for a custom solution. I closed the pull request. Specific discussions concerning the implementation (i.e. tunes) should be done in dedicated tickets.

Cheers

Markus

miweberSMI:
Hi @mfasDa,

great, thanks for confirmation!

Cheers,
Michael

PR #125, opened 2018-04-18
Fixes for ROOT6

mfasDa:
Some small stuff (missing forward declaration, incorrect forward declaration). In Config.C consequently the ROOT5 code was executed under ROOT6 and vice versa. In case of doubts that the ROOT_VERISON_CODE macro works under ROOT5 in interpreted mode (what one can deduce from the comments) I recomment to use the "__CLING__" macro -  this has been verified that it properly separates the ROOT5 and ROOT6 case also in interpreted mode.


chiarazampolli:
Hello @mfasDa ,
Thanks for the fixes, they look fine to me, I am going to merge them.

I am adding for their information @davidrohr and @shahor02 as they also worked on this part. 

Chiara

PR #126, opened 2018-04-20
Adding option to remove EMCal from simulation and reconstruction

miweberSMI:


miweberSMI:
One open question that I have, if we should also remove PHOS?

chiarazampolli:
Ciao Michael,
As Francesco said in the mail, there is no need to remove PHOS. 
For me the changes are fine.
Cheers,
Chiara

miweberSMI:
Hi,

OK, then I will merge it and port to production branches.

Cheers,
Michael

PR #111, opened 2018-03-12
Therminator2: preparing script and macro to handle new path formats

maciekniewielki:
I've modified the generation script to handle full xml file paths.

The Therminator2 AliGenerator macro is now taking a path parameter for more flexibility. The given path is then expanded using `AliDataFile::GetFileName()` which allows for relative paths, such as `"PWGCF/file.xml"`.

PR #127, opened 2018-04-20
 load AMPT from nightly builds

rbertens:
Hi @miweberSMI ,

We have a new (nightly) build of AliGenerators, so we can now use AMPT with the HIJING patch from there. This PR configures AMPT for AliDPG such that it will use this new tag of AliGenerators . For testing it, I think we need a new tag of AliDPG. 

The new build of AliGenerators also contains the updated version of Therminator2, so if we request a new tag of AliDPG, it's better to wait also for the updated AliDPG scripts for Therminator2 and fix the two problems in one go. Currently I cannot make a PR for this, as it would conflict with #111 . If you merge #111 , I can open a new PR for Therminator2. Otherwise I would ask @maciekniewielki  to amend #111 .

Thanks,

Redmer

### DIFF
--- a/MC/EXTRA/gen_ampt.sh
+++ b/MC/EXTRA/gen_ampt.sh
@@ -9,8 +9,7 @@ fi
 echo "running in `pwd`, writing HepMC to $1"
 
 # prepare environment
-source /cvmfs/alice.cern.ch/etc/login.sh
-eval $(alienv printenv AMPT::v1.26t7-v2.26t7-1)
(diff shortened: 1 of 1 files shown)


miweberSMI:
Dear @rbertens,

I tested this PR and it seems to work fine on lxplus with the latest AliPhysics tag and so I am merging it. I also merged #111 as you suggested, so that we can have a new PR for Therinator. how long would you need for it? So that we can decide, when to ask for a new tag (we have some urgent other changes waiting).

Thanks and cheers,
Michael

PR #137, opened 2018-05-24
 switch from nightlies to official builds

rbertens:


miweberSMI:
Hi @rbertens 

thanks a lot for this commit, I am merging it and porting it to the production branches.

Cheers,
Michael

PR #138, opened 2018-05-26
Fix "File too long" error in merge outputs

dberzano:


dberzano:
@chiarazampolli @fprino this fixes the error we've see in release validation. Given that I have put a barrier on the file length, the new `MD5()` line won't be executed on the Grid and won't therefore affect processing. I am merging in order to restart the validation right away.

PR #142, opened 2018-06-26
Starlight configuration updated to increase the MC efficiency for the…

iarsene:
… gamma-gamma process in the jpsi mass range


miweberSMI:
Dear @iarsene, all,

 I didn't test this adaption of the macro, but after conformation from @iarsene that it is working in the proper way, I am merging this PR.

Cheers,
Michael

PR #146, opened 2018-07-13
configuration for upgrade simulations for D-jets, HFE-jets and implem…

btrzecia:
Generator macro for an upgrade HF production, Hijing_HFjetUpgradePbPb, for new HF(D and el)-jet and B studies. 
kPythia6HFYellowReport option added for a new B decay channel.


miweberSMI:
Dear @btrzecia,

thanks for committing the generator macro for Run 3 HF studies. I am merging it now assuming that it is the configuration you want to run and port it to the production branch.

Thanks and cheers,
Michael

PR #149, opened 2018-08-01
Fixes for running AliDPG MC with Root 6

davidrohr:
This PR combines a couple of fixes I had to apply to run AliDPG simulation with Root 6.
I am not 100% sure about the include paths. For me, it worked to remove the "MC/" before the files, but I think one could alternatively set a ROOT_INCLUDE_PATH to the AliDPG root, no idea what is better.

### DIFF
--- a/MC/Config.C
+++ b/MC/Config.C
@@ -46,8 +46,8 @@ static Bool_t  purifyKine      = kTRUE;     // purifyKine flag
 static Bool_t  isFluka         = kFALSE;    // fluka flag
 
 #if ROOT_VERSION_CODE >= ROOT_VERSION(6,0,0)
-#include "MC/DetectorConfig.C"
-#include "MC/GeneratorConfig.C"
+#include "DetectorConfig.C"
+#include "GeneratorConfig.C"
 #endif
 
 void ProcessEnvironment();
--- a/MC/CreateSnapshot.C
+++ b/MC/CreateSnapshot.C
@@ -89,7 +89,7 @@ void CreateSnapshot(Int_t mode)
 
   TString ocdbRun3 = gSystem->Getenv("CONFIG_OCDBRUN3");
 
-#if ROOT_VERSION_CODE < ROOT_VERSION(6,0,0)
+#if ROOT_VERSION_CODE >= ROOT_VERSION(6,0,0)
   // in root5 the ROOT_VERSION_CODE is defined only in ACLic mode
 #else
   if(ocdbRun3.IsNull())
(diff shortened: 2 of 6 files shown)


fprino:
Hi David,

thanks! On the question about the "MC/" int he include instructions, I have no idea.
@dberzano, do you have an good advice?

Ciao, Francesco

MC/Config.C
```diff
@@ -46,8 +46,8 @@ static Bool_t  purifyKine      = kTRUE;     // purifyKine flag
 static Bool_t  isFluka         = kFALSE;    // fluka flag
 
 #if ROOT_VERSION_CODE >= ROOT_VERSION(6,0,0)
-#include "MC/DetectorConfig.C"
-#include "MC/GeneratorConfig.C"
+#include "DetectorConfig.C"
```
dberzano: As per [our instructions](https://alice-doc.github.io/alice-analysis-tutorial/analysis/ROOT5-to-6.html#how-can-i-call-macros-inside-macros) one should do:

```c++
R__ADD_INCLUDE_PATH("MC")
#include "DetectorConfig.C"
#include "GeneratorConfig.C"
```

I honestly don't know whether this works (maybe it only works with cling, don't know), can you check?

davidrohr: Hi Dario, that would not change anything. The
#include "DetectorConfig.C"
already works, the R__ADD_INCLUDE_PATH is not needed if you do not use the "MC/..." as include. The problem is that ROOT 5 uses gROOT->LoadMacro("$ALIDPG_ROOT/MC/...."), so it uses the absolute path.
And I just tested R__ADD_INCLUDE_PATH("$ALIDPG_ROOT..."), but it doesn't resolve the variable. #include "$ALIDPG_ROOT/MC/..." doesn't work either. I guess the preprocessor doesn't resolve environment variables at all, so we cannot do it on the preprocessor level.

Again, the current PR is working fine for me. I am just not sure whether it has any side effects, but I guess not.

What I meant originally is to set
ROOT_INCLUDE_PATH+=$ALIDPG_ROOT
in dpgsim.sh, such that ROOT searches in the $ALIDPG_ROOT, and then use
#include "MC/DetectorConfig.C".
However, I just tried that and it doesn't work either. I have no idea why. Is the include path not respected for .C files?

So my suggestion is to go with the PR as is, and I would add a comment to the includes that this was changed and one might need the full path and explain the problem. (So in case the include fails and someone checks the code, he knows immediately.)

dberzano: Can you just specify whether the current PR works _both for ROOT5 and ROOT6_?

davidrohr: The includes do not change at all for ROOT5, they are inside the #if ROOT_VERSION_CODE >= ROOT_VERSION(6,0,0) fence.
Well, the fixing of variables and typos affect ROOT5, but they have been wrong before anyway. But I didn't explicitly test with ROOT5.

dberzano:
OK thanks @davidrohr. @fprino: I am merging, and testing this right away with the cumulative fixes. Updates on [ALIROOT-7936](https://alice.its.cern.ch/jira/browse/ALIROOT-7936).

PR #159, opened 2018-08-29
Fix merge.C errors in ROOT 6

dberzano:
* Fix variable declared twice
* Fix ambiguous call to `StartAnalysis()`

### DIFF
--- a/DataProc/MergeOutputs/merge.C
+++ b/DataProc/MergeOutputs/merge.C
@@ -11,22 +11,22 @@ void merge(const char *filesToMerge="FilterEvents_Trees.root", const char *input
 
   // Set temporary compilation directory to current one
   gSystem->SetBuildDir(gSystem->pwd(), kTRUE);
-    
+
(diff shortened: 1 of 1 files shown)


dberzano:
@fprino @chiarazampolli merging this, it's important for the release validation (I am more and more impressed by how ROOT 5 never reported any problem!)

chiarazampolli:
Ciao Dario,

Thanks! I don't know why I did not see this error when testing locally the changes for ROOT6. Anyway, very good that you fixed it! Did you also tried it with ROOT5? Just to be sure that we don't break prodcutions (as we now tag from master, as you know).

Cheers,

Chiara

dberzano:
Yes, in any case bear in mind that the release validation also tests AliDPG master, so we would spot problems immediately.

chiarazampolli:
I know, but we don't do a RelVal every time we tag AliDPG, so not every AliDPG tag is checked...

Chiara

PR #161, opened 2018-09-01
Fix NULL vs. "" handling in PHOS CaloCells macros

dberzano:
Task `AliAnalysisTaskCaloCellsQA` exists with two different signatures. We need
to know which one we are using. Note that checking the AliRoot version is weak
(since we may not have a variable holding the version, for instance during
development). We now:

* check the macro file for the presence of the old signature,
* define a preprocessor directive accordingly (via `gInterpreter`).

The directive is processed before we invoke the corresponding
`main_QAtrain_duo.C` macro, and expanded properly there, in both ROOT 5 and 6.

Note that the old strategies are wrong/faulty for different reasons:

* conditionally checking the value of a boolean _at runtime_ does not prevent
  ROOT 6 from still processing the lines it won't actually execute, leading to
  errors for the incorrect signatures,
* loading the correct signature based on the ROOT version is wrong, since the
  change has been made in AliPhysics and it was already in use with the new
  signature for AliRoot ROOT 5-based versions

### DIFF
--- a/QA/QAtrain_duo.C
+++ b/QA/QAtrain_duo.C
@@ -68,7 +68,22 @@ void QAtrain_duo(const char *suffix="", Int_t run = 0,
   gROOT->LoadMacro("$ALICE_PHYSICS/PWGPP/HMPID/AddTaskHmpidQA.C");
   gROOT->LoadMacro("$ALICE_PHYSICS/PWGPP/T0/AddTaskT0QA.C");
   gROOT->LoadMacro("$ALICE_PHYSICS/PWGLF/FORWARD/analysis2/AddTaskForwardQA.C");
+
+  // The following macro exists across ROOT versions with two different signatures. We need to know
+  // which one we are using. Note that checking the AliRoot version is weak (since we may not have
+  // a variable holding the version, for instance during development). We ask the interpreter to
+  // process a define directive. The directive is processed before we invoke the corresponding
+  // main_QAtrain_duo.C macro, and expanded properly there, in both ROOT 5 and 6
+  if (gSystem->Exec("grep -q 'char* fname' $ALICE_PHYSICS/PWGGA/PHOSTasks/CaloCellQA/macros/AddTaskCaloCellsQA.C") == 0) {
+    // Using the old version (with char* fname, accepting NULL as value)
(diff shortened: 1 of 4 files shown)


fprino:
Hi Dario,

thanks for fixing this behaviour.
We have also another case in the QA macros in which we need to check the AliPhysics version.
This is a protection for a task that had a missing protection for a division by zero in some AliRoot version, causing crashed due to FPE.
In such case, we could not check the signature, and the only possibility is to cheeck the AliPhysics version and disable the task in case of need.
Let me know if you think that this is not OK and we should find a different protection for that task.

Ciao, Francesco

dberzano:
Checking for versions is generally a very weak approach. We don't want to rely on environment variables or internals that are different on a development environment. So, the general rule is to never check for version numbers, but for the specific functionality to be available, instead.

Possible approach: one should (IMHO) [ask forgiveness, not permission](https://stackoverflow.com/questions/12265451/ask-forgiveness-not-permission-explain), that is: I try to invoke the first version, and if it fails with a "not found error", then I try the second version.

But checking this "not found error" is easy in Python, and difficult (or, possibly, with an unstable API) in ROOT. This is why I have opted for such bizarre approach, where I actually grepped inside a file to know what to do. It's ugly, but does the job and since it greps for the "old" signature (not the new) it will always work in the future.

chiarazampolli:
Ciao Dario,

I think Francesco refers to the "disableESDtrackQA" flag here, which, if I am not wrong, you did not touch.

Cheers,

Chiara

fprino:
Hi,

yes, exactly.
I had seen that Dario did not touch it, and I was asking him if it is accpetable for this "switch" to  stay with the check on the aliroot version.
My point is that for this switch we cannot use a check on the arguments of the AddTask, not check a "no found error" when launching the configuration, since the FPE occurs when the task is run, not in the execution of the AddTask macro.
In case of development environment, the ALIEN_JDL_PACKAGES variable is not defined and the disableESDtrackQA will not be touche, i.e. the task is executed, which I think is OK, since nomally a development environment does not use the oldish aliroot version in which the task was crashing.

Cheers, Francesco

dberzano:
OK, thanks for clarifying! I did not touch it as it didn't appear to be problematic in the ROOT 6 migration context. My stance stands for future modifications 🙂

PR #163, opened 2018-09-01
Add script to run CompareAllHistos.C

dberzano:
We also allow CompareAllHistos.C to read files from any source instead of just
supporting local files by using `TFile::Open()` instead of `new TFile()`

@chiarazampolli @fprino

### DIFF
--- a/Utils/CompareAllHistos.C
+++ b/Utils/CompareAllHistos.C
@@ -66,7 +66,7 @@ void CompareAllHistos(TString filename="QAresults_v5-09-35.root", TString filena
 
 void ProcessFile(TString fname, TString dirToAnalyse){
 
-  TFile *fileBase=new TFile(fname.Data());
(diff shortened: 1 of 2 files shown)


Utils/runCompareAllHistos.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash -e
```
dberzano: This is a steering script used by the release validation

Utils/CompareAllHistos.C
```diff
@@ -66,7 +66,7 @@ void CompareAllHistos(TString filename="QAresults_v5-09-35.root", TString filena
 
 void ProcessFile(TString fname, TString dirToAnalyse){
 
-  TFile *fileBase=new TFile(fname.Data());
+  TFile *fileBase=TFile::Open(fname.Data());
```
dberzano: This change allows to read files from remote locations as well

PR #168, opened 2018-09-03
Add TPC spline task in raw data QA train

fprino:


QA/main_QAtrain_duo.C
```diff
@@ -929,6 +937,11 @@ void ProcessEnvironmentVars()
     printf("Set doPIDqa=%d according to environment variables\n",doPIDqa);
   }
   //
+  if ( !(envS=gSystem->Getenv("TPCsplines")).IsNull() && CheckEnvS("TPCsplines",envS) ) {
```
dberzano: @fprino: I see references to new environment variables. Should we export those somewhere? Is this supposed to run on the Grid?

fprino: Hi Dario,

the environment variable is used just in case one wants to override the default switches of the macro. For instance, doTPCsplines is by default kTRUE in the macro, but it can be switched off without modifying the macro by defining the environment variable TPCsplines and setting its value to 0.
If we run with the default, no exporting of variable is needed.

This is the same structure that we have also for other tasks.

Ciao, Francesco

dberzano: OK thanks for clarifying! To me this PR is fine.

PR #172, opened 2018-09-10
a new config for generating PYTHIA based HF jet (by D. Godoy)

sashingo:


sashingo:
Commit for the MC request of HF jet production in pp at 13 TeV (ALIROOT-7974)  
https://alice.its.cern.ch/jira/browse/ALIROOT-7974

PR #173, opened 2018-09-10
Adding single muon parametrizations

miweberSMI:


miweberSMI:
Dear all,

we are still waiting for confirmation from @pillot about the initial tests of this macro, but I am merging already now, so that we have it in the new tag and could start in case the green light comes from him. If changes are needed they can then go to a new tag.

Cheers,
Michael

pillot:
Hi Micheal,

I am afraid it was too early. I already did several modifications and I am currently testing them...

Cheers,
Philippe

miweberSMI:
Hi @pillot,

OK, don't worry. Then we will wait for your changes to be committed before asking a new tag, how long do you need?

Thanks and cheers,
Michael

PR #174, opened 2018-09-10
load choosen macro + prepare future options + add comments

pillot:
Hi @miweberSMI , all,

I kept the same philosophy but I changed a bit the order, to select the macro to be used according to the option and load only this one. The configuration macros now need to implement the same function CreateGenerator() that is used in the generic macro.

I also added the possibility to activate some specific setup in view of adding other macros. Typically for Jpsi radiative decay one needs to load EvtGen.

I changed the name of the configuration macro to make it more explicit. There should not be much versions of the same macro with this naming.

I also added several comments to describe all this.

I think it should also work with ROOT6 but I tested it only with ROOT5. 

Cheers,
Philippe


miweberSMI:
Hi @pillot ,

thanks a lot for your quick and nice work on the single muon macro, it looks very good to me. At the moment I can also not test it with ROOT6, but we should be able to do that during validation.
I am merging it for now, so that we can start the production with a new tag (perhaps you can already modify the description of the production ticket accrodingly?).

Thanks and cheers,
Michael

PR #175, opened 2018-09-11
Set different vmem ulimit on ROOT 5/6

dberzano:


dberzano:
@chiarazampolli @fprino @davidrohr @pzhristov this is the problem that was plaguing us. So simple to fix. Didn't think to check it unfortunately.

I am merging it for tests (ROOT 5 processing still gets the same limit, so no regression on prod). Please speak up in case of problems and we can revert it or do it better.

chiarazampolli:
Ciao Dario,

Thanks!
Unfortunately, we cannot have 4GB on the grid to my knowledge. So we should find out why ROOT6 needs more. It looks like an offset at the beginning of the processing that makes the difference.

Cheers,

Chiara

pzhristov:
Hi Chiara,

We are talking about virtual memory and it is an very old limit that is not maintained anymore on most of the sites (as far as I know). I think it is very bad idea to limit ourselves the max. virtual memory - most of the jobs will anyway finish.

Concerning the Root6 - we had about 100 MB overhead because of the precompiled headers. As soon as we have in place the clang modules (by the end of the year), we will have less memory consumption compared to Root5.

Cheers,
Peter

chiarazampolli:
Dear Peter,

I definitely understand your point. The limit was put very long ago, when we had issues with jobs killed die to too high virtual memory usage. The numbers that are typically quoted are 3.5 for the VM and 2.5 for the resident, even though it is true that jobs might even survive when they use 8 GB. I think we added this limit to be sure to fit all sites, it was something we did in agreement with Latchezar and Ruben. If it is obsolete, we can remove it for good. 

Thanks a lot for the information about the overhead!

Cheers,

Chiara

PR #176, opened 2018-09-11
Disable vmem ulimit for ROOT 6

dberzano:


dberzano:
vmem limit is kept for ROOT 5, but eliminated for ROOT 6 processing as discussed with @pzhristov.

cc @chiarazampolli

chiarazampolli:
Ciao Dario,

Thanks! I will check with Ruben if we should keep this for ROOT5.

Cheers,

Chiara

PR #178, opened 2018-09-12
add missing include path + compile parameterization macro locally

pillot:
Hi @miweberSMI ,

I used a symbolic link as suggested (with a generic name, which is easier to handle in the generic macro). I tested and it works for me.

Cheers,
Philippe


miweberSMI:
Dear @pillot,

thanks a lot for this update. I checked the full simulation chain and it runs w/o errors. PR is merged now and should appear in the next tag.

Cheers,
Michael

PR #181, opened 2018-09-17
Don't run the CheckESD for now.

chiarazampolli:
We need to converge on how to set the PID response in the macro.
Since currently the output of this macro is not looked at and not
even saved, it is ok to not run it.
See also discussion in https://github.com/alisw/AliDPG/commit/52bc50b3b690e4265406d8518a91e6adcccb5067#diff-110caea4b0dcf9bdb2a94892627e82f9


chiarazampolli:
Francesco checked this PR, and he found it ok, so I am merging it now.

Chiara

PR #192, opened 2018-10-04
LoadLibraries is not needed

chiarazampolli:


chiarazampolli:
Hello @shahor02,

I removed the "LoadLibraries" call also from the TPC SP calibration part. I tested it, and everything works, in the sense that the processing was successful - I did not check the results. Do you agree to merge it?

Cheers,

Chiara

fprino:
HI Chiara, all,

fine with me.

Francesco

chiarazampolli:
Many thanks to both Francesco and Ruben, merging now.

BTW, should I remove completely the macro? Now it is used nowhere.

Chiara

shahor02:
I would keep it for the moment.
Cheers,
 Ruben

chiarazampolli:
Ok, I thought the same, and then was doubtful as one can always recover it from git.

Merging like this now.

Chiara

PR #193, opened 2018-10-09
Update help function, changing default mode to NULL

miweberSMI:


miweberSMI:
Dear all,

I updated the help function, it should now contain all available options. I also updated the TWiki page accordingly: https://twiki.cern.ch/twiki/bin/view/ALICE/AliDPGMonteCarloTutorial 
In order to print it also if no argument is provided I changed the default mode to NULL (was OCDB,full before).

Let me know what you think or if I am missing something.

Cheers,
Michael

chiarazampolli:
Ciao Michael,

Many thanks, I think it is fine like this! In case we notice that something is still missing, we can add it. 
I also agree on the default for the "mode".

Please, go on with merging.

Cheers,

Chiara

miweberSMI:
Hi @chiarazampolli,

thanks for checking. I am merging this PR.

Cheers,
Michael

PR #197, opened 2018-10-17
Script to compare two tags.

chiarazampolli:
There is also the script runCompareAllHistos.sh which was written to
be integrated in the RelVal, this is what I use.

### DIFF
--- a/Utils/compareTags.sh
+++ b/Utils/compareTags.sh
@@ -1,0 +1,37 @@
+#!/bin/bash -e
+
+set -x
+
+LATEST_RELCANDIDATE_PATH=$1
+REFERE_RELCANDIDATE_PATH=$2
+
+for filename in QAresults_barrel.root QAresults.root QAresults_AOD.root 
+do
+    
+    if [ "$filename" = "QAresults_AOD.root" ]
+    then
+	directory=pass1
+	for folder in PIDqa MultSelection CheckVertexAOD CheckAODTracks PWG3_D2H_QA
+	do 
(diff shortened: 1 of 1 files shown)


chiarazampolli:
This is just a new script for Release Validation. I will merge it.

fprino:
Hi,

with this version of the script there are some TDirectories of QAresults.root that are not checked, e.g.
Vertex_Performance, SPD_Performance, T0_Performance and many others, sice you run only on TDitrectories whose name contains one of these strings: TPC TOF ITS TRD CaloQA_default CaloQA_trigEMC.
An option could be to call CompareAllHistos.C with \\"\\" as third argument, so that all TDirectories in the root file are checked.

Ciao, Francesco

chiarazampolli:
Ciao Francesco,

Thanks! I will check to modify the script, but usually this is what I check (ITS, TRD, TOF, TPC, AOD, EMCAL) in the RelVal... The script as it is is also creating an output log file per directory, for the sake of simplicity to isolate them. I will see if I can make it like this in a smart way.

Cheers,

Chiara

fprino:
Hi,

OK. Just keep in mind that with the string selection "ITS" you are not checks all the ITS plots, because the  subdetector QA has directories named SPD_Performance, SDD_Performance...
A possibility could be to add SDD SPD SSD and ertex to the list of directories in the loop.

II do not see issues in having a single file with the output of all directories, but if you wish I can modify the macro such that it writes the output in different files grouped per detector.

Cheers, Francesco

PR #200, opened 2018-10-29
Ideal vertex in case of generator only production + adding new HF cus…

miweberSMI:
…tom macro

### DIFF
--- a/MC/CustomGenerators/PWGHF/Hijing_Xe_HF_Omegaccc_bkg.C
+++ b/MC/CustomGenerators/PWGHF/Hijing_Xe_HF_Omegaccc_bkg.C
@@ -1,0 +1,29 @@
+AliGenerator *
+GeneratorCustom(TString opt = "")
+{
+
+    AliPDG::AddParticlesToPdgDataBase();
+
+    AliGenCocktail *ctl  = GeneratorCocktail("Hijing_HF");
+  
+        AliGenerator   *hij  = GeneratorHijing();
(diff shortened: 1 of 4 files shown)


miweberSMI:
Dear @afestant, @chiarazampolli, @fprino, 

I finally managed (lack of time so far, apologies) to modify the central macros to distinguish two cases: ideal vertex (0,0,0) in case of generator only productions and CDB vertex if full simulation. It is not beautiful, but it works. What do you think? 
In addition, I committed also the custom macro. @afestant is something else missing (other macros that were modified)?

Cheers,
Michael

afestant:
Dear @miweberSMI,

thanks a lot for the commit. I didn't even modify the SimulationConfig.C because in case you set "--GeneratorOnly", sim.SetRunGeneratorOnly(kTRUE); is called and AliSimulation::RunGeneratorOnly doesn't use the vertex from the OCDB if I have understood correctly. In any case, the commit looks fine to me.

Thanks again
Cheers,
Andrea

miweberSMI:
Hi @afestant,

good point! I reverted the changes in SimulationConfig.C since it is not needed.
Should we merge it now, @chiarazampolli, @fprino? Then we could do some testing on the LS4+ production once a new tag is there.

Thanks and cheers,
Michael

chiarazampolli:
Ciao Michael,
Yes, for me we can merge.
Cheers,
Chiara

miweberSMI:
OK, thanks. I will merge then.

PR #201, opened 2018-10-31
Use all triggers in cpass1 QA for Pb-Pb

fprino:


fprino:
Hi,

this is the commit that enables the QA for all triggers in CPass1.
I used the variable ALIEN_JDL_LPMPASSNAME, which accordding to JDL files is the one containing the pass type information.
@catalinristea can you please confirm that this is the right variable to use?

Cheers, Francesco

catalinristea:
Ciao @fprino ,

This (ALIEN_JDL_)LPMPASSNAME is set automatically, ie we don't manage ourselves. 
We're safe using it, as all the production tags are relying on it. 
You did the correct if clause, since apart from cpass1_pass1 we could have at some point also cpass1_pass2, and it will work.
Please merge to master.

Thanks a lot!

Yours,
Catalin.

chiarazampolli:
Hello,

Thanks to both, I merged! Tomorrow I will make the new tag, please, let me know if something else is needed.

Cheers,

Chiara

PR #203, opened 2018-11-09
new generators for Upsilon(nS) to dimuon in pPb and Pbp at 8.16TeV

pillot:


miweberSMI:
Hi @pillot 

thanks for this PR! 
According to our discussion, I am merging it now.

Cheers,
Michael

PR #211, opened 2018-11-19
Incorporating the check for the integrity of the ResTree in the script

chiarazampolli:
Hello @shahor02 ,

Can you check if these changes are fine with you?

Chiara

### DIFF
--- a/DataProc/CPass0/checkResTree.C
+++ b/DataProc/CPass0/checkResTree.C
@@ -9,10 +9,18 @@ void checkResTree(const char* fname = "ResidualTrees.root", const char* tname = 
   TTree* tr = (TTree*)tf.Get(tname); // we want macro to crash if the tree is not there
   int nev = tr->GetEntries();
   int rdb = 0;
+  bool corrupted = kFALSE;
   for (int i=0;i<nev;i++) {
     rdb = tr->GetEntry(i);
     if (rdb<=0 || rdb>10000) {
       printf("CORRUPTED Residuals file, ev %d got %d bytes\n",i,rdb);
+      corrupted = kTRUE;
     } 
+  }
+  if (corrupted) {
+    printf("The residual tree file is corrupted\n");
+  }
+  else {
+    printf("The residual tree file is OK\n");
   }
 }
--- a/DataProc/CPass0/checkResTree.C
+++ b/DataProc/CPass0/checkResTree.C
@@ -13,8 +13,9 @@ void checkResTree(const char* fname = "ResidualTrees.root", const char* tname = 
   for (int i=0;i<nev;i++) {
     rdb = tr->GetEntry(i);
     if (rdb<=0 || rdb>10000) {
-      printf("CORRUPTED Residuals file, ev %d got %d bytes\n",i,rdb);
+      printf("CORRUPTED Residuals file, ev %d got %d bytes, job to be Killed\n",i,rdb);
       corrupted = kTRUE;
+      break;
     } 
   }
   if (corrupted) {
--- a/DataProc/CPass0/checkResTree.C
+++ b/DataProc/CPass0/checkResTree.C
@@ -13,15 +13,10 @@ void checkResTree(const char* fname = "ResidualTrees.root", const char* tname = 
   for (int i=0;i<nev;i++) {
     rdb = tr->GetEntry(i);
     if (rdb<=0 || rdb>10000) {
-      printf("CORRUPTED Residuals file, ev %d got %d bytes, job to be Killed\n",i,rdb);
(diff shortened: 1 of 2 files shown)


DataProc/CPass0/checkResTree.C
```diff
@@ -9,10 +9,18 @@ void checkResTree(const char* fname = "ResidualTrees.root", const char* tname =
...
   int rdb = 0;
+  bool corrupted = kFALSE;
   for (int i=0;i<nev;i++) {
     rdb = tr->GetEntry(i);
     if (rdb<=0 || rdb>10000) {
       printf("CORRUPTED Residuals file, ev %d got %d bytes\n",i,rdb);
+      corrupted = kTRUE;
     } 
   }
+  if (corrupted) {
```
shahor02: If the tree is corrupted, it is not guaranteed to go through the all ```nev``` of the loop, if the messages of this ```if .. else ..``` in the log are to be used by the validation, it is better to add a ```break;``` after the ```corrupted = kTRUE;```

DataProc/CPass0/runCPass0.sh
```diff
@@ -336,6 +336,32 @@ if [ $exitcode -ne 0 ]; then
...
+echo ""
+echo "running the following checkResTree.C macro:"
+cat checkResTree.C
+echo ""
+echo executing aliroot -l -b -q -x "checkResTree.C()"
+echo "" >&2
+echo "checkResTree.C" >&2
+timeStart=`date +%s`
+time aliroot -l -b -q -x "checkResTree.C" &>> integrity.log
+exitcode=$?
```
shahor02: The ``integrity.log`` will contain the status, but will the validation check it? I guess this log has to be added to the ``/alice/validation/cpass/0/validation.rc``, but should not one also add phrase to grep to the ``validation.sh``? 
Alternatively, one can changed the output in case of the corruption to 
``printf("The residual tree file is corrupted, the job must be Killed\n");``
then this will be intercepted by the current ``validation.sh``. 
I have locally one corrupted chunk, will test it now.

DataProc/CPass0/checkResTree.C
chiarazampolli: Actually I added this only to have a "positive" message in the log if the file is not corrupted.

DataProc/CPass0/runCPass0.sh
chiarazampolli: The validation.rc lists the files that have to be there, not the ones that are checked. The validation.sh search for the specified strings in all lig files (but the syswatch and tag.log, see https://alimonitor.cern.ch/users/download.jsp?view=true&path=/alice/validation/validation.sh).

chiarazampolli: Ah, and yes, I would add CORRUPTED to the list of strings to check. If you change it to "Killed" it will already work.

DataProc/CPass0/checkResTree.C
shahor02: I've checked with the script as you have it and the macro as
```
void checkResTree(const char* fname = "ResidualTrees.root", const char* tname = "delta")
{
  TFile tf(fname);
  TTree* tr = (TTree*)tf.Get(tname); // we want macro to crash if the tree is not there
  int nev = tr->GetEntries();
  int rdb = 0;
  bool corrupted = kFALSE;
  for (int i=0;i<nev;i++) {
    rdb = tr->GetEntry(i);
    if (rdb<=0 || rdb>10000) {
      printf("CORRUPTED Residuals file, ev %d got %d bytes, to be Killed\n",i,rdb);
      corrupted = kTRUE;
      break;
    } 
  }
  if (corrupted) {
    printf("The residual tree file is corrupted: job must be Killed\n");
  }
  else {
    printf("The residual tree file is OK\n");
  }
}
```
using existing ```/alice/validation/validation.sh```, it works both on the corrupted and on healthy ```ResidualTrees.root``` as expected.

chiarazampolli: Ok, so only the break was added, right? I can add it sure.

shahor02: Also added "Killed" to messages signaling a corruption

chiarazampolli: Ok, done. Can you check? If you think it is fine, I will merge the changes.
Thanks!

chiarazampolli:
Hi Ruben, 

Can you check again after the latest changes?

Chiara

shahor02:
Checked, works as planned.

chiarazampolli:
Ok, great, merging. I will tag after the CB.

PR #214, opened 2018-11-19
 implement decoding of mass and charges for magnetic monopoles via the PDG code

sebaleh:


miweberSMI:
Hi @chiarazampolli, @fprino ,

here are some small changes for the magnetic monopole simulations. 
Before the pdgconfig (--pdg XXXXXX) was set to 60000000, now one can set also charges and mass with this number (--pdg 6abcdefg) if it is between 60000000 and 70000000. Since we are touching the general GEANT4 config for that, I think it might be good to have somebody else checking as well.

Thanks and cheers,
Michael

fprino:
Hi Michael,

to me it looks OK.
I guess it was tested locally before committing, right?

Ciao, Francesco

miweberSMI:
Hi @fprino 

thanks for your review. Yes, we tested:
- pions (PDG=211) stay unchanged 
- monopoles (PDG>60000000&&PDG<70000000) can be set with the logic as explained in the comments and give reasonable results
Then I am merging it into the master.

Thanks and cheers,
Michael

PR #217, opened 2018-11-27
Do not update MeanVertex in CPass0 also for 18r

chiarazampolli:
Hello @afestant , 

I ping you only as I don't know Davide's git account... (If you do, please add him). 
I implemented the check also for 18r, now, for MeanVertex to not be done in CPass0 for this period. 

Cheers,

Chiara


afestant:
Hi @chiarazampolli,

thanks a lot for implementing the fix.

Cheers,
Andrea

PR #223, opened 2018-12-21
Add macro skimming raw events including TPC data for EJ/G1 triggers f…

mfasDa:
…rom the CENT cluster

### DIFF
--- a/DataProc/skimming/skimRawCENTEJ1.C
+++ b/DataProc/skimming/skimRawCENTEJ1.C
@@ -1,0 +1,377 @@
+/////////////////////////////////////////////
+// macro for skimming RAW data
+// Used for CENT cluster EJ1 triggers from PbPb 2018 data
+// 21.11.2018
+// Martin Poghosyan (adaption Markus Fasel including 
+// central barrel detectors for EJ1/EG1-triggered events in CENT(NOPMD))
+/////////////////////////////////////////////
+
+#include <bitset>
+#include <TROOT.h>
+#include <TError.h>
+#include <TFile.h>
+#include <TTree.h>
+#include <TChain.h>
+#include <TGrid.h>
+#include <Riostream.h>
+#include <TEnv.h>
+#include <TAlienCollection.h>
+#include "AliRawReader.h"
+
+#include "event.h"
+#include "AliRawVEvent.h"
+#include "AliRawEventHeaderBase.h"
+#include "AliRawVEquipment.h"
+#include "AliRawEquipmentHeader.h"
+#include "AliRawDataHeader.h"
+#include "AliRawData.h"
+#include "AliDAQ.h"
+#include "AliCDBManager.h"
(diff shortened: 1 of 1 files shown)


DataProc/skimming/skimRawCENTEJ1.C
```diff
@@ -0,0 +1,377 @@
...
+    1  /* 26 --- */
+  };
+
+
+  // select branches	
+  Bool_t kDetAccept[AliDAQ::kNDetectors+1] = {
+    kTRUE , //SPD
+    kFALSE, //SDD
+    kTRUE , //SSD
+    kTRUE,  //TPC
```
shahor02: @mfasDa we were running in HLT-C mode, therefore the TPC branches will be empty (still ok to have it).

DataProc/skimming/skimRawCENTEJ1.C
```diff
@@ -0,0 +1,377 @@
...
+    kTRUE , //ZDC
+    kFALSE, //ACORDE
+    kTRUE , //TRG
+    kTRUE , //EMCAL
+    kFALSE, //DAQ_TEST
+    kFALSE, //EMPTY
+    kFALSE, //AD
+    kFALSE, //MFT
+    kFALSE, //FIT
+    kFALSE, //HLT
```
shahor02: you need to enable HLT branch to get TPC clusters

mfasDa: Hi @shahor02 ,

Thanks for the comments - I modified the script accordingly. Please have a look!

Thanks!

Cheers

Markus

shahor02: I see that now you disabled the TPC. While it will be empty, I think it is better to have it also, unless you tested that the TPC reconstruction works with this branch disabled.

chiarazampolli:
Hello,

What about this PR? should it be merged?

Chiara

mfasDa:
Hi @chiarazampolli ,

The last comment from @shahor02 was addressed (re-enabling TPC branch), so from my side the PR is ready.

Cheers

Markus

chiarazampolli:
Hello Markus,

Thanks!

This macro was still not used in production, right? While the one SkimRaw is the one we used for http://alimonitor.cern.ch/production/raw.jsp?jt_field1=skim, correct?

Chiara

mfasDa:
Hello Chiara,

Yes. The purpose is to skim for a very small amount of data (i.e. 1 long run or a few shorter runs) only the EMCAL triggers but from the CENT cluster, including the TPC data, in order to develop the ITS-TRD-TOF tracking in a high local track density environment, where it might be essential to have the TPC tracks in order to tune cuts/parameters. 

Cheers

Markus

chiarazampolli:
Ciao Markus,

Ok, thanks, merging now.

Chiara

PR #224, opened 2019-01-07
RELDIS via AliDPG

rbertens:


miweberSMI:
Hi @rbertens,

I tested now this PR and found two small things:
L1737: probably the description needs to be changed here to "RELDIS" (not so important)
L1740: you are missing a "()" here (that is important ;) )
After changing this it was running fine (did not check the outputs though, but the Kinematics file was filled).

Thanks and cheers,
Michael

rbertens:
@miweberSMI 
oops - thanks for spotting! Since I couldn't start the simulation to begin with, these bugs slipped through the cracks ... PR has been amended, let me know if it's OK now.
my best,
redmer

miweberSMI:
Hi @rbertens,

thanks a lot for your quick action, looks good to me now. I am merging this PR.

Cheers,
Michael

PR #225, opened 2019-01-09
Adding more processing options for RELDIS:

miweberSMI:
- first event to be processed
- tracking neutrons or spectators only


miweberSMI:
Hi @rbertens,

do you agree with the changes? If yes, I would merge that PR...

Thanks and cheers,
Michael

miweberSMI:
Dear all,

I am merging now this PR, if there is a problem spotted later on, we can always update.

Cheers,
Michael

PR #229, opened 2019-01-31
New single muons parametrisation for PbPb 5TeV (LHC18r)

obugnon:
Update of the single muons parametrisation for PbPb at 5 TeV from tunning on LHC18r data

### DIFF
--- a/MC/CustomGenerators/PWGDQ/Muon_GenParamSingle_PbPb5TeV_1.C
+++ b/MC/CustomGenerators/PWGDQ/Muon_GenParamSingle_PbPb5TeV_1.C
@@ -2,7 +2,7 @@
 // Parameterized generator for single muons //
 //==========================================//
 // Info about this particular tunning:      //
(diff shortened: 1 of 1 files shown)


miweberSMI:
Hi @obugnon,

I trust that the cahnges made here are correct. I was wondering if one should create a new one and leave the one for 2015 unchanged? But perhaps @pillot who is the original author of the macro can comment as well. 

Cheers,
Michael

pillot:
Hi @miweberSMI,

I think in this case we can update the parameters. Since the tuning is done on the same collision system at the same energy with the same event and track selections, we expect it to be identical, and indeed the differences are very small. So we can consider it as a little improvement (the stat is higher so the tuning is more precise) of the same tuning. If we had to repeat the MC for 2015, we would use this new tuning as well.

Cheers,
Philippe

miweberSMI:
Hi @pillot,

thanks for your reply, then I will merge the commit to the master.

Cheers,
Michael

PR #231, opened 2019-02-06
auto-disable ADA PMT shielding if year<2017

ecalvovi:
Update DetectorInit(...) to automatically disable ADA PMT shielding if the year is less than 2017 as this is the year when the shielding was installed.

### DIFF
--- a/MC/DetectorConfig.C
+++ b/MC/DetectorConfig.C
@@ -540,6 +540,7 @@ DetectorInit(Int_t tag)
      if (!isFluka){ // does not work for now with Fluka
         //=================== AD parameters ============================
         AliAD *AD = new AliADv1("AD", "normal AD");
+	if(year<2017) ((AliADv1*) AD)->EnablePmtShieldingADA(kFALSE);
         if( tag == kDetectorPhosOnly)
         AD->DisableStepManager();
      }


catalinristea:
Hi @fprino , @chiarazampolli , @miweberSMI,

This feature is new in AliRoot, please see
https://github.com/alisw/AliRoot/commit/6e4bd2bb6ac6716e49aa6415fe643394398a36bb#diff-2f93983eb1bd59ecbeb60528628288a7

We'll need a version protection in AliDPG, as right now it will crash the upcoming LF Nuclei production from https://alice.its.cern.ch/jira/browse/ALIROOT-8197

Thanks!

Yours,
Catalin.

fprino:
Hi,

yes, we need the protection.
We can do it based on AliPhysics version, which is however not the cleanest possible solution.
Another cleaner way would be to do it using the class version of AliAD.
When looking the the change in AliRoot:
https://github.com/alisw/AliRoot/commit/6e4bd2bb6ac6716e49aa6415fe643394398a36bb
I saw tat two new data members were added.
However, in this commit the ClassDef was not updated.
This need to be fixed in AliRoot (@ecalvovi I can do it if you agree).
Then, we could implement a simple check like:

      if (!isFluka){ // does not work for now with Fluka
         //=================== AD parameters ============================
         AliAD *AD = new AliADv1("AD", "normal AD");
-       if(year<2017) ((AliADv1*) AD)->EnablePmtShieldingADA(kFALSE);
+       if(year<2017 && AD->Class_Version()>=2) ((AliADv1*) AD)->EnablePmtShieldingADA(kFALSE);
         if( tag == kDetectorPhosOnly)
         AD->DisableStepManager();
      }

This modification in AliDPG I can commit it already now: it will prevent the crash in case we use an old AliRoot with new AliDPG.

Cheers, Francesco

fprino:
Hi,

another (and maybe cleaner) solution could be to set the data member fEnabledPmtShldADA to kFALSE in  the constructor of AliADv1, based either on the content of the title, that is passed as second argument to the constructor itself.
In this way, we would not have issues of coupling between the AliDPG and AliRoot versions, by doing something like
```
AliAD *AD = 0x0;
if (year<2017) AD = new AliADv1("AD", "normal AD 2015");
else AD = new AliADv1("AD", "normal AD");
```
and then AliADv1 could manage to set accordingly the configuration by parsing the title.

Cheers, Francesco

fprino:
Hi,

while we wait for the feedback from the AD experts on the two solutions that I proposed, I will temporarily revert this commit, to allow us to have an AliDPG tag with some other modifications that are needed for upcoming productions

Best regards, Francesco

ecalvovi:
Hi @fprino, 

I like your second solution. I will implement it. 

Cheers,
Erneso.

fprino:
HI Ernesto,

OK, thanks!
Maybe, while doing the implementation, it is worth that you check that modifying the title of the class does not have side effects.

Cheers, Francesco

ecalvovi:
Hi Francesco,

Now the title passed to the AD constructor will be used to extract the year.
Valid format for title is: "ignored tokens year=2017 other ignored tokens"
Example: 
AliAD *AD = 0x0;
AD = new AliADv1("AD", Form("normal AD year=%d", year));

A new member ParseYear() called only in the constructor accepting name and title will search in the title 
for the regular expression "year=YYYY" and if found then the year will be used to decide whether to construct the ADA PMT shielding or not.
if the regex is not found, or the constructor without name-title is used, the ADA PMT shielding is constructed by default. 
In the later case you can still use the method  EnablePmtShieldingADA(const Bool_t flag) to alter this behavior.

The pull request is here: https://github.com/alisw/AliRoot/pull/902

Cheers,
Ernesto

fprino:
Hi Ernesto,

thanks. Looks very good to me.
If I got it right, in case the year is not found in the title (which is what will happen when we will have to run productions with old AliDPG), the simulation won't crash in AD, but it will keep the default value for the  EnablePmtShieldingADA. 
If this is the case, I'd say that we can move on in this direction and you can commit the modification in AliDPG, to define the title based on the year.

Thanks!

Cheers, Francesco

ecalvovi:
Hi Francesco,

Yes, this is the case, if no info about the year in the title then default values are assumed (and by default AD will construct the shielding to reflect the most recent change to geometry).
The pull request is here: https://github.com/alisw/AliDPG/pull/239

Thanks and cheers,
Ernesto.

PR #232, opened 2019-02-06
Add new box generator for GEANT3/GEANT4 studies in LF

mpuccio:
Ciao @miweberSMI ,

this new generator has been requested by some analysers to check GEANT4 response for nuclei in the MC. I did few checks locally and it was working. We will ask a small production (we still remember that we owe you a QA for the embedding :) ) using this generator in the next days.

Thanks a lot,
Max


miweberSMI:
Hi @mpuccio,

I just checked on lxplus and at least the simulation stage finishes w/o errors. So I think everything is fine here on the generator side (I just checked GEANT3 though).

Cheers,
Michael

PR #233, opened 2019-02-07
Modifications to D2H custom generators for Lc

jezwilkinson:
Modify existing custom generator for Lc in pPb for signal training in ML studies to include option for forced decay chain of kaon from Lc (see https://github.com/alisw/AliRoot/pull/893). Also added new generator for PbPb that includes same.

### DIFF
--- a/MC/CustomGenerators/PWGHF/Hijing_PbPb_Lc.C
+++ b/MC/CustomGenerators/PWGHF/Hijing_PbPb_Lc.C
@@ -1,0 +1,69 @@
+AliGenerator *
+GeneratorCustom(TString opt = "")
+{
+  AliGenCocktail *ctl  = GeneratorCocktail("Hijing_HF");
+  AliGenerator   *hij  = GeneratorHijing();
+  ctl->AddGenerator(hij, "Hijing", 1.);
+
+  const Char_t *label[2][4] = {
+    "chadr PYTHIA", "chadr PYTHIA", "cele PYTHIA", "chadr PYTHIA", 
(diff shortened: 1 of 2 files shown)


chiarazampolli:
This was already tested, this is why I merged it.

Chiara

chiarazampolli:
Ciao Jeremy,

Sorry, I reverted your commit as I had not noticed that you had implemented a new "Hijing_HF001.C" file. Why did you not use the same as we changed for the test? I mean: why do you want to keep all possibility + the Lc in a separate file, and not just everything in the same? 

Chiara

jezwilkinson:
Ciao Chiara,
I modified the Hijing_HF001.C we used for the test and added it as "Hijing_PbPb_Lc.C" - modification was so that the same macro can also be used for the forced pKpi decay in PbPb if needed, and it seems more organised to keep the macros that are specific to the BDT training separate from the ones we use for "general" productions (since we also handled the pPb production macro in the same way).

chiarazampolli:
Ciao Jeremy,

But then, I would keep the Lc --> K0sp and Lc --> pKpi in the "standard" file, and have in the new (to be renamed, with a "BDT" suffix and no "Lc") only the part for the BDT MC. What do you think?

BTW, pPb had to go to a different file as the weighting for centrality (for example) is not used there.

Cheers,

Chiara

chiarazampolli:
Ciao Jeremy,

Sorry to bother, but we will tag most likely quite soon, so I was wondering whether you agree with my proposal for this macro. 

Thanks and cheers,

Chiara

PR #237, opened 2019-02-11
Generators for dedicated Lc production with forced kaon decay

jezwilkinson:
As https://github.com/alisw/AliDPG/pull/233/, but with PbPb methods integrated into existing Hijing_HF001.C as requested by @chiarazampolli .


chiarazampolli:
Ciao @jezwilkinson , 

Thanks for the changes. So now also the BDT is in the main macro - I thought we wanted it separate. 

Anyway, for me it is fine, I think that the PWGHF conveners should approve. Do you know their github accounts? 

Cheers,

Chiara

chiarazampolli:
Dear @arossi81 , @agrelli,

Can you review too these changes, and in case, add Xiaoming?

Thanks,

Chiara

arossi81:
Hi @chiarazampolli,@jezwilkinson,
at first look it's ok for me. I have two comments:
- is the trigger particle pdg number charge sensitive? I guess so. If you confirm it: I know that we did the same for pp and p-Pb, but do we want to generate only Lc+? We could randomise the generation of Lc+ and Lc-.
- remove, close to the pKpi dedicated entry, the comment "//Force Lc decay mode in PYTHIA to Kc --> K0s+p with forced K0->K0S->pi+pi-  
Cheers,
Andrea

jezwilkinson:
Hi @arossi81,
I don't believe the SetTriggerParticle is charge-sensitive; the trigger is handled by https://github.com/alisw/AliRoot/blob/9e5bb12171e8782c17446eb0faf6a4b40b78c9bb/PYTHIA6/AliPythia6/AliGenPythia.cxx#L1154 where it uses "if(TMath::Abs(pdg)==fTriggerParticle)" to decide. So here there shouldn't be an issue.
For the comments, thanks for spotting, I've fixed them in a new commit to the branch.
Cheers,
--Jeremy

arossi81:
+1
Hi @jezwilkinson,
thanks!
Cheers,
Andrea

chiarazampolli:
Thanks @arossi81 @jezwilkinson , merging.

PR #239, opened 2019-02-15
Decide if create ADA PMT shielding based on year

ecalvovi:
Decide whether to create or not ADA PMT shielding based on the year passed to the title of the ADv1 constructor.
If year<=0, then do not pass that into the title, in which case AD will be constructed with default values (by default AD assumes that it must construct the shielding).

### DIFF
--- a/MC/DetectorConfig.C
+++ b/MC/DetectorConfig.C
@@ -539,7 +539,11 @@ DetectorInit(Int_t tag)
   if (iAD){
      if (!isFluka){ // does not work for now with Fluka
         //=================== AD parameters ============================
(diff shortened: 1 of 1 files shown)


fprino:
Hi Ernesto,
thanks!
Looks OK to me.
@catalinristea , @chiarazampolli , @miweberSMI , do you agree to merge it?

Cheers, Francesco

chiarazampolli:
Ciao, 

Looks ok to me too.

Chiara

fprino:
Hi,

Thanks Chiara for the feedback. I'll merge now the commit, so that we can run some tests with the master of AliDPG.

Cheers, Francesco

PR #241, opened 2019-02-21
Run AMPT_v226t7 in standalone (needed for CPU idle issue)

miweberSMI:


miweberSMI:
Hi @chiarazampolli,

thanks for your quick feedback!
I had this first but the point is that we call also the generator like that to distinguish it from the "old" (Run1) AMPT. We could update this as well though. Do you have a good suggestion for a name?

Cheers,
Michael

chiarazampolli:
Ciao Michael,

Ah, ok, I thought you wanted to use this approach for all AMPT versions from now on. 
Do we have any Run1 AMPT done in AliDPG? I mean, do we need to have that case? Otherwise, could not just "AMPT" be fine, and work with all versions?

Chiara

miweberSMI:
Hi @chiara,

my point is that there are two generators in AliDPG now:
- AMPT
- AMPT_v226t7
which are both also named like that.
They differ quite a lot since the second is steered via AliGenerators and much more flexible (and we can easily adapt new versions). I don't know if we have productions in AliDPG with the first. 
If you agree, I will merge this PR now and later change the name of the second to "AMPT_v2" (this would have the advantage that also running it with "--generator AMPT_v226t7" would be recognized).

Cheers,
Michael

chiarazampolli:
Ciao Michael, 

Sure, for me it is fine, thanks!

Chiara

PR #242, opened 2019-02-21
Fix after changes in return value of the AddTaskEmcalTriggerQA_QAtrain.C

chiarazampolli:
In AliPhysics v5-09-44b and from v5-09-46, the returned object of
the macro changed. Since we don't use it, we just remove it, to
remain backward compatible without checks on the version of AliPhysics.


chiarazampolli:
@mfasDa , FYI as discussed in ALIROOT-6204.

PR #243, opened 2019-02-21
Change Y range for Lc generation (in K0s+p)

chiarazampolli:


chiarazampolli:
Dear @arossi81 , @agrelli, @andreaalici,
Here the changes we discussed for the range in y to generate the Lc. Note that I changed only the Lc --> K0s+p decay channel, I don't know if you want the same for the other. 
Cheers,
Chiara

chiarazampolli:
Ciao,

Closing, there is already #246 for this, and here I made the mistake in modifying the wrong "y" range (sorry). 

Chiara

PR #245, opened 2019-02-21
Adapt naming of AMPT generator v2:

miweberSMI:
- as discussed on https://github.com/alisw/AliDPG/pull/241
- to make clear that not only v226t7 will be done like that but all versions onwards


miweberSMI:
Ciao Chiara,

thanks a lot, I am merging then...

Cheers,
Michael

PR #247, opened 2019-02-26
Adding possibility to set from outside the pTHardMin for HIJING

chiarazampolli:


chiarazampolli:
Dear @fprino , @miweberSMI , 

Can you have a look at the proposed change, to be able to set the cut on the pthard for HIJING from outside?

Cheers,

Chiara

MC/Config.C
```diff
@@ -279,6 +281,11 @@ ProcessEnvironment()
     abort();
   }
 
+  // ptHardMin for HIJING simulation
+  ptHardMinHijing = 3.;
```
miweberSMI: move to default 2.3 ?

MC/Config.C
```diff
@@ -24,6 +24,7 @@ static Int_t triggerConfig   = 0.;        // trigger
 static Int_t   pdgConfig       = 0;         // PDG value 
 static Float_t bminConfig      = 0.;        // impact parameter min
 static Float_t bmaxConfig      = 20.;       // impact parameter max
+static Float_t ptHardMinHijing = 3.;        // min cut on ptHard for Hijing simulations
```
miweberSMI: move to default 2.3 ?

chiarazampolli: Ciao Michael,

I thought that since 2.3 is proven to be not ok, maybe for now we can set the default to 3. I think we should decide on the number when we have the configuration finalized... 

But if you prefer, I put this to 2.3, to be compatible to current value. I will do it.

Chiara

miweberSMI:
Hi @chiarazampolli ,

thanks a lot for this! My proposal (see single comments in two lines) would be to move the default to 2.3 to reproduce the old setting when this option is not used?
On the other hand, if we stay with 3 it would be annoying to always add this option, so probably to stay with 3 as default.

MC/Config.C
chiarazampolli: See below :-)

MC/Config.C
miweberSMI: Hi Chiara,

as I wrote in my comment, I think for both there is justification. Actually, thinking about it again, I would keep the 3 as default. We just need to remember if we ever do Run-1 productions again ;) 

Cheers,
Michael

chiarazampolli:
Ciao Michael,

Let's keep 2.3 for now, we will change this once we find a better value.

Cheers,

Chiara

miweberSMI:
OK with me.

chiarazampolli:
Ciao Michael,

Let's do 2.3 for now. In case we decide to go for 3, we will change it for the new MC.
Anyway, we can do the tests to validate this number with the tarball... 

Chiara

chiarazampolli:
Ciao Michael,

Thanks! Let's see if @fprino has further comments. 
I will be offline at some point till ~14:30, so if you all agree, you can then "squash and merge", or I will do it after lunch, so that we can start some tests on the grid. 

Chiara

chiarazampolli:
I am now merging to start the tests on the grid (it might be that Francesco is offline, travelling). In case there is any problem, we can revert (and anyway, we will not tag).

Chiara

fprino:
Hi,

sorry I was offline between P2 and meetings.
All fine with me. Thanks for implementing it.

Cheers, Francesco

chiarazampolli:
Ciao Francesco,

Thanks! And sorry for now waiting your feedback, I thought you might have been driving home. 

Chiara

PR #248, opened 2019-02-27
Use HLT TPC-compression-only mode by default

davidrohr:
This needs current AliRoot master, i.e. commits b52ecaa4477312144a9611dffb47057ebc659e77 and b52ecaa4477312144a9611dffb47057ebc659e77.

It changes the HLT behavior such by default only the TPC compression runs, which should speed up MC simulation. We should double-check that this has no side effects.

### DIFF
--- a/MC/SimulationConfig.C
+++ b/MC/SimulationConfig.C
@@ -17,6 +17,8 @@
 #include "TROOT.h"
 #endif
 
+#include "../Utils/CheckAliRootVersion.C"
+
 enum ESimulation_t {
   kSimulationDefault,
   kSimulationMuon,
@@ -230,6 +232,7 @@ void SimulationDefault(AliSimulation &sim)
   //
   // HLT settings
   TString hltConfig = "auto";
+  if (IsAliPhysicsMoreRecentThanOrEqualTo("v5-09-46")) hltConfig = "auto chains=TPC-compression-only";
   if (gSystem->Getenv("CONFIG_HLT"))
     hltConfig = gSystem->Getenv("CONFIG_HLT");
   sim.SetRunHLT(hltConfig.Data());


chiarazampolli:
Ciao David,

Do you mean b52ecaa4477312144a9611dffb47057ebc659e77 and  f866ffff6b32cbd3d358b4db1aaba0430ab30f23?

What will happen if we run this with an old AliRoot? Since new AliDPG may run with old AliRoot, we might need to put a protection on the AliRoot version for this patch, right?

Cheers,

Chiara

davidrohr:
Ciao Chiara,

yes, I meant these. Sorry for the copy and paste error.

it cannot work with the old AliRoot, as the new HLT chain we use does not exist in old AliRoot. So it should throw a fatal error when you run with the old AliRoot version.
Then we'll have to add a protection for the AliRoot version, or make this somehow settable.

Just FYI: The background of this patch and the ones to AliRoot is that Peter and Andreas asked me why HLT takes so much time in the simulation and this patch should fix it. If you want to use it for MC simulation with an old AliRoot, the patches must be backported. But first, it would be good if someone confirms that the HLT compute time is good now.

Cheers
David

chiarazampolli:
Ciao David,

Thanks for the feedback. Yes, I thought the changes were related to the comments by Andreas (he posted one on the G4 productions related to HLT). 

I think we can include a protection in terms of AliRoot version (see e.g. https://github.com/alisw/AliDPG/blob/master/QA/main_QAtrain_duo.C, from line 218). @fprino , @miweberSMI , do you agree? Back-porting is also possible, but we need to know how much we would gain. Also, it would then risk to be a bit complicated, as we have several aliroot versions that we use for 2015, 2016, 2017, 2018 data... Maybe we should first check how much we gain (overall, not just in simulation), and decide. 

As for checking, did you run a test yourself? Anyway, we can also run one (either on the grid, or privately - we will discuss in the DPG meeting today). 

Cheers,

Chiara

davidrohr:
Ciao Chiara,

I ran a test locally, and verified that the HLT clusters are available during TPC reconstruction. Also, HLT reconstruction is much faster as it runs only the compression now, which gets rid of the tpc transformation map.
Still, the cluster finding emulation and also compression in HLT are not so efficient, so they might still take some time. But if we want to have compatible clusters in MC; there is nothing more that we can do (except for a general improvement of the old HWCF emulation and compression code). But it shouldn't be the 30% anymore that Andreas was claiming.

Cheers
David

chiarazampolli:
Ciao David,

This was 30% of the overall simulation?

Chiara

davidrohr:
That is what I had understood

chiarazampolli:
Ciao David, 

There is no way to get (e.g.) the list of chains in the HLT, and check if the new one is present, in order to apply your change instead of looking at the AliRoot version, right?

Chiara

davidrohr:
Ciao Chiara,

no, i don't think that is possible. Of cource we could add some kind of flag by adding something to a header? But I am wondering:this is certainly not the first case we added a feature to aliroot which is then used by AliDPG? How do you usually check whether a feature is available? If there is nothing, i think we should invent something general, which can also be used for o2 later.
Kind Regards
David Rohr

(Sent from my mobile, excuse the typos.)

chiarazampolli:
Ciao David,
Typically we use the AliRoot version from an environment variable defined in the job. We were looking for something independent from it, but since it does not exist, we will do it this way. 

Chiara

chiarazampolli:
PS: before merging, I will run some tests, and anyway, we will first need this in a tagged AliRoot.

davidrohr:
Ciao Chiara,

for the long run, we should discuss within DPG to handle this.
A very simple solution, which will directly be compatible to more projects like O2, would be to have a simple text-file with list of features, which is copied during installation.
AliDPG could just grep for the required feature. This would also work well with backported features.

chiarazampolli:
Ciao David,

Yes, for the long run we might use a different approach, for now we will use the tag.
Now I am running some tests, to see the impact of your changes (100 events). 
Note that we can accept this PR only one we have a new AliRoot tag.

Cheers,

Chiara

chiarazampolli:
Ciao David,
How do I check if the chain simulated is the correct one? I don't see a specific printout (for now) in the sim.log, and since I ran 2x... I wanted to be sure the correct was taken.
Thanks,
Chiara

davidrohr:
Ciao Chiara,
HLT would usually print the chain it runs, but perhaps its at Info level and suppressed in the normal simulation.
You can check for the presence of messages wrt. GlobalEsdConverter and TPCCAGlobalMerger, which would indicate that the tracking is running.
You can also send me the two log files and I can have a look.

chiarazampolli:
Ciao David, 

Found it, it should be fine. I will anyway send you the logs when done.

Thanks!

Chiara

davidrohr:
@chiarazampolli : Do we still need this PR, or did you implement the feature with an AliRoot version check in the meantime?

chiarazampolli:
Ciao David,

Yes, we need to implement the check on the AliRoot version still. So either you do this in your PR, or we merge it and immediately add this check. 
Sorry, but I might not manage to do so today... 

Cheers,

Chiara

davidrohr:
Hi @chiarazampolli : I just had a look at https://github.com/alisw/AliDPG/blob/master/QA/main_QAtrain_duo.C, to implement the version check in the same way.
However, there is quite some string parsing involved.

Probably it is better if we add a general function for checking AliRoot / AliPhysics version, which contains the string parsing?
Something like bool CheckAliPhysicsVersion(int v1, int v2, int v3) {...} and then we would just use
if (CheckAliPhysicsVersion(5, 9, 47)) {...}

chiarazampolli:
Ciao David,
Definitely this would be the best. 
If you want, we can implement this function, and then merge your PR using it. 
Cheers,
Chiara

chiarazampolli:
Ciao @davidrohr , 

Can you check if you can integrate the macro here: https://github.com/alisw/AliDPG/pull/256 in your PR?

Cheers,

Chiara

davidrohr:
@chiarazampolli : The function seems to implement a > comparison. Isn't >= what we need?
I'll have to check for 5-09-46, so with the current function, I'll have to write:
IsAliPhysicsMoreRecentThan("v5-09-45") (It seems to ignore the a, b, c, etc suffices, so this will check for 5-09-46.

davidrohr:
Updated this PR with the AliPhysics version check

chiarazampolli:
Ciao David,

Maybe indeed we should add ">=". @fprino , what do you think? 
The "a,b,c" are ignored, for now, yes. At present, for the cases we have, it does not matter, but it might if we introduce something in a "patched" version (those with the letters). @fprino , should we introduce this too? 
Note that in the past there was a case (by "mistake") when tag 20e (I think) was done from 20c and not 20d... So the ordering in the letter suffix but not be always respected (even if I know about one case only).

Chiara

chiarazampolli:
Or, instead of the ">=", you can call the two functions, and put them in OR, but it would then do the parsing 2x... 

Chiara

davidrohr:
Well, logically, if a new feature is introduced at version x, I want to check that I am running at least version X. So I think the >= is the comparison used most often, so I'd make this the default.

chiarazampolli:
What about this?

https://github.com/alisw/AliDPG/pull/257

It is like a merging of the conditions between the existing functions. Nothing else.

fprino:
Hi,

sorry for the low reply.
Yes, indeed if we need to check >=, the solution of PR 257 is OK.

Ciao, Francesco

davidrohr:
Updated to use #257

PR #249, opened 2019-02-28
First version of HM Pythia6 generator

miweberSMI:


miweberSMI:
Hi @valzacc,

I tested this macro, but got the following error:

Users/physics/aliceSW/AliDPG/MC/CustomGenerators/PWGMM/Pythia6_HighMult.C:8:15: error: no member named 'SetTriggerChargedMultiplicity' in 'AliGenerator'
      pythia->SetTriggerChargedMultiplicity(60, 1.4) //  multiplicity threshold set to 60 in |eta| < 1.4

Am I missing something?

Thanks and cheers,
Michael

fprino:
Hi,

I think the problem comes from the fact that SetTriggerChargedMultiplicity is a method of AliGenPythia, which is not inherited from AliGenerator.
So, you probably need to cast the return value of GeneratorPythia6 into AliGenPythia.

Cheers, Francesco

miweberSMI:
Hi @fprino 

you are right that did the trick (see update). Probably this is not detected by ROOT5 and was therefore not spotted in the tests before?

Cheers,
Michael

valzacc:
Yes, I tested with ROOT5 in cvmfs and I did not get any error. Thanks for your tests and PR!
Ciao,
Valentina

PR #251, opened 2019-03-04
config for Hijing+Pythia HF-jet Pb-Pb productions

btrzecia:
Config for HF-jet productions in Pb-Pb. Hijing+Pythia6 (c/b enhancement with 10 Pythia per Hijing event).


miweberSMI:
Hi @btrzecia ,

I tested this macro and it is running fine. I still will wait before merging, since I understand that you would like to do also embedding with it.

Cheers,
Michael

miweberSMI:
Hi @btrzecia,

thanks for the update for embedding. Looks OK to me. I am merging now...

Cheers,
Michael

PR #256, opened 2019-03-18
Macro to compare AliPhysics versions

fprino:


chiarazampolli:
Ciao Francesco,

Thanks for the implementation! 
I was wondering: for the function "IsEqual", should we return kFALSE if the packages are not defined? Probably it is the most common case that you are running with a different version.
BTW, I think we don't have such cases at present, when we need perfect match, do we? But I agree it is good to include it too. 

Pinging @davidrohr , as this is related to https://github.com/alisw/AliDPG/pull/248.

Chiara

fprino:
Hi Chiara,

thanks for the feedback.
I have changed the return code for the method Is Equal as you suggested.
I agree with you that at present we do not need the method IsEqual, but I thought that it does not hurt to have it.

Ciao, Francesco

chiarazampolli:
Ciao Francesco,

Thanks to you! and I agree that it could turn to be useful the IsEqual.

For me we can merge.

Chiara

PR #257, opened 2019-03-20
Adding MoreRecentThanOrEqualTo case

chiarazampolli:


davidrohr:
Isn't this still ">" not ">="?

chiarazampolli:
mmmh... If they are euqal it returns kTRUE, otherwise it checks if it is ">", no?
Line 71:
if(ver==vvv && n1==nnn1 && n2==nnn2) return kTRUE;

davidrohr:
Ah, I missed that line.
I'd just change `if(n2<=nnn2) return kFALSE;` to `if(n2<nnn2) return kFALSE;`, which I think is more intuitive and shorter than the additional check.

chiarazampolli:
Ok, I can change like that. 
But I have an issue: I cannot test it:

root [0] gROOT->LoadMacro("Utils/CheckAliRootVersion.C")
W-TEnvRec::ChangeValue: duplicate entry <Library.vector<short>=vector.dll vectorbool.dll> for level 0; ignored
(Int_t)0
root [1] IsAliPhysicsEqualTo("v5-09-45")
Error: sscanf parameter mismatch param[0] C u Utils/CheckAliRootVersion.C:91:
*** Interpreter error recovered ***
root [2]

davidrohr:
I guess it should be sscanf(version.Data() (also in the other function)

chiarazampolli:
Yes. 
Committed now, fixed and modified (and tested).
@fprino , if you and David agree, I would merge, then he could use the new function, and we could merge also his PR248.
Chiara

fprino:
Hi,

yes, I agree. Thanks!

Ciao, Francesco

chiarazampolli:
Thanks, merging now. Then @davidrohr , you can change in your PR, and finally we will merge it :-) Sorry for the late action about this... 
Chiara

PR #258, opened 2019-03-29
Updated AOD macro to keep logs, intended for central productions

catalinristea:
Ciao @chiarazampolli , @fprino 

Here is the pull request to save the logs for re-filtering productions - please merge if you find it appropriate.
Thanks!

Yours,
Catalin.

### DIFF
--- a/AOD/main_AODtrainRawAndMC.C
+++ b/AOD/main_AODtrainRawAndMC.C
@@ -116,7 +116,7 @@ Int_t run_numbers[10] = {177580}; // Set the run range, for testing
  *            Refiltering settings                *
  **************************************************/
 
-TString     aliphysics_version = "v5-09-17c-01-1"; // *CHANGE ME IF MORE RECENT IN GRID*
+TString     aliphysics_version = "v5-09-46-01-1"; // *CHANGE ME IF MORE RECENT IN GRID*
 
 Int_t       iAODanalysis        = 0;      // AOD based analysis 
 // Maximum number of files per job (gives size of AOD)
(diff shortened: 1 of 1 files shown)


catalinristea:
Ciao Chiara,

Well, the AliPhysics changed just because I committed the AMPT version for AOD217.

We're running locally to generate grid files, so the version is hardcoded and gets into the created JDL. The workflow is different from standard, meaning this version is going into JDL, not the other way around.

In general it will work if we update afterwards the version in JDL (and it will be the only place where we actually define the version), it will only matter if the tasks or parameters are changing from a version to the other, this is why better to use the same version as tested locally when we generate the JDL. So I would say it's not necessary to parametrize it, as usually the ALIEN_JDL_PACKAGES is not defined locally.

Thanks!

Yours,
Catalin.

chiarazampolli:
Hello Catalin,

Indeed, I think you already told me that for the refiltering you use really the plugin and not an LPM interface. 
But then, my question is: does it mean that recent refilterings were done with the previous AliPhysics? I think we already used v5-09-46... E.g. for muon_calo pass3 of 18r.

Cheers,

Chiara

catalinristea:
Ciao Chiara,

Indeed, you are right, this is how it goes.

What you see in ML is what we ran - when we have a new refiltering, I take the macro from AliDPG and update by hand the AliPhysics version, then I generate locally the configuration. To be sure, I also copy the AOD macro in grid, along with the generating log and environment file used locally.

Thanks!

Yours,
Catalin.

chiarazampolli:
Ok, clear now.
For me we can merge, if @fprino agrees too.

Chiara

fprino:
HI,

yes, I agree. I am merging now.
Thanks!

Ciao, Francesco

PR #259, opened 2019-04-02
Printing the content of the BKG directory in the sim.log

chiarazampolli:


chiarazampolli:
Hello @catalinristea , @fprino , @miweberSMI ,
This is to print in the sim.log the content of the background folder, so that we can check its size. 
I am merging, it should be harmless... But please revert if you see any issue.

Chiara

PR #261, opened 2019-04-09
MC update for ALIDPG_ROOT when running with tar ball, relevant only f…

catalinristea:
…or embedding

Ciao @chiarazampolli, @fprino 

Here is the update to make ALIDPG_ROOT as absolute path - relevant only for MC embedding productions and only when running with AliDPG tar ball.

Thanks!

Yours,
Catalin.


chiarazampolli:
Fine with me! Thanks Catalin for taking care!

PR #263, opened 2019-04-17
Modifications for proper handling of AliMultSelection task

miweberSMI:


miweberSMI:
Hi @chiarazampolli, @fprino, and @catalinristea ,

I removed the "RemoveBKGFromGAlice" part in the extract embedded wrapper.
My question is if there are other things needed?
- The mixed mode?
- keeping of unfiltered galice and Kinematics (at the moment we do it only in debug mode)?
I marked them also in the code...

Thanks and cheers,
Michael

miweberSMI:
hi @chiarazampolli,

thanks for your feedback. I hope I applied your suggested changes properly. I am testing on lxplus right now.

Cheers,
Michael

miweberSMI:
Hi all again,

so the test revealed one more problem, which is fixed with the last commit to this PR (the check on filtered galice file needs to be removed of course). After that the full production chain is running for embedding with and w/o signal filtering. At least I get all the files without errors, but I did not check the content. If you all agree, we could merge from my side...

Cheers,
Michael

chiarazampolli:
Thanks Michael! I remember that when I tried I also had to do something with the galice files, but I did not do it with a lot of care, and was simply commenting out some lines :-)
I agree to merge!

PR #265, opened 2019-04-19
Pythia8 + inj strangeness for pp 5TeV 2017

bianchil:
New custom generator for flat-pt injected lambda, Xi and Omega into Pythia8 Monash 2013.
Will be used for pp5TeV 2017 anchored MC production to increase stats for strange baryons


miweberSMI:
Hi @bianchil, 

I tested the generator macro and simulation and reconstruction run fine. I am assuming that the macro is also producing what you would like to have. Could you (or conveners) add also the JIRA ticket of the production, where this is needed? Or it doesn't exist yet?

Thanks and cheers,
Michael

miweberSMI:
Dear all,

I am merging now, since we got the OK from conveners.

Cheers,
Michael

PR #271, opened 2019-06-26
friend info for extracted tracks is also filtered

shahor02:
Filtered friend tree is stored by default in AliESDfriend_EMB.root


chiarazampolli:
Thanks, Ruben!
Chiara

PR #273, opened 2019-07-08
Embedding capabilities for Starlight macro

miweberSMI:


iarsene:
Hi Michael,
As I commented also in the JIRA ticket (https://alice.its.cern.ch/jira/browse/ALIROOT-8277?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&focusedCommentId=232660), the main issue is with identifying the 2 electrons from the 
gamma+gamma process of the Starlight generator. In the past productions we used the fact that these 2 electrons are the first 2 tracks in stack. If we switch and have Hijing first, we can't use this trick anymore. However, I think we can access the generator information and we should be able to disentangle them.
For the other processes there should be no problem.
Cheers, ionut

miweberSMI:
Dear all,

Ionut approved the change in ALIROOT-8277, so I am merging this PR.

Cheers,
Michael

PR #274, opened 2019-07-08
Fallback to builtin event selection

mfasDa:
Do to issues with the centrality selection the AliEventSelection
cannot be used in automatic mode in the QA train. Therefore the
fallback to the builtin event selection is used on the QA train.


mfasDa:
@chiarazampolli @fprino This enables the fallback using the builtin event selection as discussed.

chiarazampolli:
Thanks! Merging now.
Chiara

PR #276, opened 2019-07-10
removed mixed mode (from sh script only)

miweberSMI:


miweberSMI:
Hi @chiarazampolli , @fprino , @shahor02,

for safety, I removed the mixed mode from the dpg script. Could you have a look? 
At the moment I am keeping the option in the macros, or should we remove it completely? What do you think?

Cheers,
Michael

shahor02:
Hi @miweberSMI 

I think suppressing it only in the macro is enough, I would keep it in the macro in case somebody will need to study the crosstalk between the clusters/tracks of underlying and embedded events.

Cheers,
 Ruben

chiarazampolli:
Hello,
I agree with @shahor02 , let's remove it from the script, but leave it in the macro. 
Thanks!
Chiara

miweberSMI:
OK, thx. Then I am merging now...

PR #280, opened 2019-07-15
Re-add muon refit task

hupereir:
- re-added AliAnalysisTaskMuonRefit
- use automatic loading of alignment storage to setup the task

### DIFF
--- a/AOD/AODtrainRawAndMC.C
+++ b/AOD/AODtrainRawAndMC.C
@@ -9,6 +9,7 @@ void AODtrainRawAndMC(Int_t merge=0, Bool_t isMC=kFALSE, Bool_t refilteringMode=
   gROOT->LoadMacro("$ALICE_PHYSICS/PWGPP/PilotTrain/AddTaskCDBconnect.C");
   gROOT->LoadMacro("$ALICE_PHYSICS/OADB/macros/AddTaskPhysicsSelection.C");
   gROOT->LoadMacro("$ALICE_PHYSICS/PWGPP/macros/AddTaskFilteredTree.C");
+  gROOT->LoadMacro("$ALICE_PHYSICS/PWG/muondep/AddTaskMuonRefit.C");
   gROOT->LoadMacro("$ALICE_PHYSICS/OADB/COMMON/MULTIPLICITY/macros/AddTaskMultSelection.C");
   gROOT->LoadMacro("$ALICE_PHYSICS/OADB/macros/AddTaskCentrality.C");
   gROOT->LoadMacro("$ALICE_PHYSICS/PWGLF/FORWARD/analysis2/AddTaskForwardMult.C");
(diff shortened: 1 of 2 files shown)


hupereir:
untested.
based on the macro at https://alice.its.cern.ch/jira/browse/ALIROOT-6646

chiarazampolli:
Ciao @hupereir , 

Thanks! 
If we leave the setting to "1", the task will be always called. Is it ok? Also for PPass, for example (and then I would ask: it is a time/CPU consuming task?)

Cheers,

Chiara

hupereir:
> Ciao @hupereir ,
> 
> Thanks!
> If we leave the setting to "1", the task will be always called. Is it ok?

Probably not a good idea no !
So leave it to zero by default, and make sure that it is set to 1 manually only for productions that require it ? 

> Also for PPass, for example (and then I would ask: it is a time/CPU consuming task?)
> 
> Cheers,
> 
> Chiara

chiarazampolli:
Then we need a way to set it from outside, as AliDPG is tagged... 
@catalinristea , do you want to add a JDL variable for this?

fprino:
Hi,

can we add a switch based on an environment variable that we can set ad-hoc in the JDL in those cases in which we want to execute the task?

Ciao, Francesco

chiarazampolli:
Ciao Francesco, 

Yes, this is what I meant. 
@catalinristea , do you prefer that I take care, or will you?

Cheers,

Chiara

catalinristea:
Ciao Chiara,

This cannot be controlled by the JDL, as the configuration is generated by AliAnalysisAlien running locally - all automatic generated files (macro, task configurations saved in root file, validation, JDL, etc) are copied and executed, as they are, in grid. Probably we could alter all these files if we want, but this could be error prone. To conclude, I'll change the variable locally, before generating configuration, to have it for this re-filtering.

If whole re-filtering needs to be changed, we can start thinking to it - to run for example with AliDPG. This is valid also for the AOD re-filtering for 2010/2011, vertexingHF.

Thanks!

Yours,
Catalin.

chiarazampolli:
Hello Catalin,

Ok, so we should have the setting = 0 in master, then for the 18q/r muon-calo pass3 we can use a local modified version of it. In this way we can start as soon as this change is in place (@hupereir , can you take care?). For the AOD 2010 and 2011 refiltering, I would propose to either do the same, or wait to discuss this on Wed at the DPG meeting. What do you and @fprino  think?

Cheers,

Chiara

catalinristea:
Ciao Chiara,

Sure, agreed!
Thanks.

Yours,
Catalin.

fprino:
OK also for me.
Thanks! Ciao, Francesco

chiarazampolli:
Thanks! Merging now.
Chiara

PR #281, opened 2019-07-17
Fixing iteration over tasks (ALIROOT-8287)

mfasDa:


chiarazampolli:
Ciao @mfasDa , 

Thanks! I was about to commit the same fix. 

One suggestion: there is a lot of printouts like:
"W-AliEmcalTriggerQATask::RetrieveEventObjects: AliEmcalTriggerQATask_CINT7: Could not retrieve centrality information! Assuming 99"
Are they really needed? We have one per task for every event. 

This of course is in case a change in AliPhysics. So if you consider doing it, we should redo the RelVal with updated AliPhysics. 

Cheers,

Chiara

mfasDa:
Ciao @chiarazampolli,

Thanks for merging!

The warning for the missing multiplicity object needs to be fixed in AliPhysics. Could be done, although for user tasks I prefer to have the option for warning about the centrality available. I will prepare a fix!

Cheers

Markus

chiarazampolli:
Ciao,

Yes, I know it is for AliPhysics, so I would have waited for this fix to redo the RelVal, but ok, let's just use the updated AliDPG. 

Thanks!

Chiara

PR #268, opened 2019-05-22
update macros for PbPb 2018 MC simulations

ffionda:
The commit includes:
1) custom generator macro for PbPb 2018 (Hijing_Jpsiee002.C ) with centrality dependent injection (thanks to Ionut for the suggestion)
2) GenJPsiParaSet.C to set up user pT / y parametrizations (thanks to Markus for providing pT shapes from PbPb 2015)
3) small change in GeneratorConfig.C in order to use the functions at point 2)

PR #284, opened 2019-07-18
Fixes for ROOT6

miweberSMI:


miweberSMI:
Hi @chiarazampolli , @fprino,

I added two fixes:
- "Data()" was missing for geant physics list (not sure, why this was no problem so far)
- I fixed the problems for ROOT6 that were introduced in https://github.com/alisw/AliDPG/pull/268
Note that the latter means that Generator_Jpsiee using UserParams cannot be run with ROOT6, thsi will need another fix by the PWG. But at least the GeneratorConfig macro can be called now again with ROOT6.

Cheers,
Michael

chiarazampolli:
Ciao,

The "Data()" is related to the new feature of the Physics List for G4, that I introduced lately. Sorry for forgetting this. 

Chiara

miweberSMI:
Dear @chiarazampolli,

thanks for your feedback, I will merge then...

Cheers,
Michael

PR #285, opened 2019-07-18
Merge gen.headers in RemoveBKGFromGalice

shahor02:
Hi @miweberSMI @chiarazampolli 
This PR merges the gen.headers of the BG event to that of the embedded one, rewriting the TE tree in the galice.root (written to galice_EMB.root) of the latter when RemoveBKGFromGalice is called.

Note, now the RemoveBKGFromGalice is called automatically from the ExtractEmbedded() with {{keepMixed==false}}.

The steering script will need to save  galice_EMB.root as galice.root, as it is done for AliESDs.

### DIFF
--- a/MC/ExtractEmbedded.C
+++ b/MC/ExtractEmbedded.C
@@ -32,7 +32,8 @@
 #include "TTree.h"
 #include "TClonesArray.h"
 #include "TString.h"
-
+#include "AliGenCocktailEventHeader.h"
+#include "AliHeader.h"
 #include "AliESDfriend.h"
 #include "AliESDVZEROfriend.h"
 #include "AliESDTZEROfriend.h"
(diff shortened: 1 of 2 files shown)


miweberSMI:
Hi @shahor02,

thanks for your commit. 
Just to understand: your comment means we need to uncomment these lines in dpgsim.sh again:

	## Removed for proper handling of AliMultSelection task
	#mv "galice.root" "galice_ORIG.root"
	#mv "galice_EMB.root" "galice.root"
    else
	mv "AliESDs_EMB.root" "AliESDs.root"
	mv "AliESDfriends_EMB.root" "AliESDfriends.root"
       	## Removed for proper handling of AliMultSelection task
	#mv "galice_EMB.root" "galice.root"

Thanks and cheers,
Michael

shahor02:
Hi @miweberSMI, yes, with mixed=false the filtering will create ``galice_EMB.root`` with headers of BKG and injected events merged (and reference of the galice of the BKG removed), and this ``galice_EMB.root`` should be saved as "galice.root" of the filtered production.

chiarazampolli:
Hello @shahor02 , 

Could you maybe then change also dpgsim.sh, so that we have all needed changes in the same PR?

Thanks!

Chiara

miweberSMI:
Dear all,

after discussion in the DPG coordination meeting, I am merging this PR. 

Thanks and cheers,
Michael

miweberSMI:
Hi @shahor02, all,

I was testing with a latest AliDPG/AliRoot and get the following warning, when analyzing events that were created with MC-to-MC embedding and signal filtering:

W-AliMCEvent::GetTrack: AliMCEvent::GetEntry: Index out of range
W-AliMCEvent::AssignGeneratorIndex: AliMCEvent::AssignGeneratorIndex: 0-pointer to particle j       -1 npart      855 nsumpart       22 Nprimaries       22

Does any of you, where this comes from? And if we will need another fix for that issue?

Thanks and cheers,
Michael

shahor02:
Hi @miweberSMI ,
Indeed, I don't set the track multiplicity in the Hijing header to 0, and this creates problems in the AssignGeneratorIndex, since the algorithm in this case looks for the tracks of the bg. event. I can commit a fix (probably on Saturday, since now I am traveling and my internet connection is barely enough to read emails), but to be sure: isn't the multiplicity of the hijing header used in the analysis, e.g. in centrality task. Note that the total multiplicity from all headers is anyway accumulated in the coctail header, but I I am not sure which one is used.

Cheers,
 Ruben

miweberSMI:
Hi @shahor02,

ah OK, then this will need more thinking. In any case, there is no hurry. So, I will first discuss it in the DPG meeting and perhaps come back to you for advice.

Thanks a lot and cheers,
Michael

PR #287, opened 2019-07-23
Typo fixed to use option for G4 physics lists

chiarazampolli:


chiarazampolli:
This fixes the fact that the G4 tests (LHC19g3b, c, d) were all using the INCLXX lists (even if they should have not).

PR #293, opened 2019-08-02
Add new macro for Lc / D_s dedicated generation, including HM trigger options

jezwilkinson:
proposed to be used by multiple productions for analyses aiming at QM: for dedicated productions of Lc and D_s, including sign of trigger particle and option for HMSPD/V0HM triggers

### DIFF
--- a/MC/CustomGenerators/PWGHF/Pythia6_Perugia2011_D2Hdedicated.C
+++ b/MC/CustomGenerators/PWGHF/Pythia6_Perugia2011_D2Hdedicated.C
@@ -4,7 +4,11 @@ AliGenerator *GeneratorCustom(TString opt = "")
 
   Decay_t decay[4] = {AliDecayer::kLcpKpi, AliDecayer::kLcpK0S, AliDecayer::kLcpK0SBDTsig, AliDecayer::kHadronicDWithV0};
   Int_t   triggerParticle[4] = {4122,  4122, 4122, 431};
-  TString optList[4] = {"LcpKpi","LcpK0S", "LcpK0SBDTsig","DsDedicated"};
+  TString optList[3][4] = {{"LcpKpi","LcpK0S", "LcpK0SBDTsig","DsDedicated"},
+			   {"LcpKpiHMSPD","LcpK0SHMSPD","LcpK0SBDTsigHMSPD","DsDedicatedHMSPD"},
+			   {"LcpKpiV0HM", "LcpK0SV0HM" ,"LcpK0SBDTsigV0HM" ,"DsDedicatedV0HM"}};
+		
+
   TString label[4] = {"LcpKpi PYTHIA","LcpK0S PYTHIA", "LcpK0S PYTHIA (forced K decay)", "DsDedicated PYTHIA"};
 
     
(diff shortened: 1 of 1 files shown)


miweberSMI:
Dear @jezwilkinson,

while the discussion on the V0HM settings continues on the mailing thread, I was already starting to test this configuration. Can I ask you if you tested on ROOT5 or ROOT6? Because on ROOT5 I get a errors: 
"Error: Symbol {"LcpKpi","LcpK0S","LcpK0SBDTsig","DsDedicated"} is not defined in current scope  /home/mweber/AliSW/AliDPG/MC/CustomGenerators/PWGHF/Pythia6_Perugia2011_D2Hdedicated.C:7:" etc.

When testing on ROOT6, this seems to work, but then I get an error due to the decayer (I am using ALIROOT master): 
"error: no member named 'kLcpKpi' in 'TVirtualMCDecayer'

Cheers,
Michael

jezwilkinson:
Hi @miweberSMI,
indeed i had had problems with running the DPGsim script itself to test on my system (but this was independent of the generator macro), but checked that the methods should work - I had tested under ROOT6 that the string assignment would work, and didn't realise that it would cause an issue under ROOT5... I suppose this will have to be reworked but I can't see a tidy way of assigning it in this case.
I'm also very surprised that it doesn't manage to find the kLcpKpi symbol, as this has existed in EVGEN/AliDecayer.h since almost two years now. Does AliDecayer need to be explicitly included somewhere for this to work?
Cheers,
--Jeremy

MC/CustomGenerators/PWGHF/Pythia6_Perugia2011_D2Hdedicated.C
```diff
@@ -0,0 +1,45 @@
+AliGenerator *GeneratorCustom(TString opt = "")
+
+{
+
+  Decay_t decay[4] = {AliDecayer::kLcpKpi, AliDecayer::kLcpK0S, AliDecayer::kLcpK0SBDTsig, AliDecayer::kHadronicDWithV0};
+  Int_t   triggerParticle[4] = {4122,  4122, 4122, 431};
+  TString optList[3][4] = {{"LcpKpi","LcpK0S", "LcpK0SBDTsig","DsDedicated"},
```
miweberSMI: Char_t* instead of TString (for ROOT5)?

miweberSMI:
Hi @jezwilkinson,

changing from TString to Char_t* would at least solve the problem for ROOT5 (which we will anyhow use to run the production).

Cheers,
Michael

miweberSMI:
Hi again,

another question/comment is on the naming of the macro(s). In HF we have now a mixture of macros that are actually now built on one another (this one is quasi an extension of Lc), perhaps the conveners could quickly comment if this is what they want... Or if some could be merged into one.

Thanks and cheers,
Michael

jezwilkinson:
Hi Michael,
Thanks for the suggestion on using Char_t*, I have committed now (apologies, I missed the update to the thread earlier), and updated the HM trigger eta range following some discussion with Francesco via mail.
Indeed this macro most likely deprecates the existing Lc one. The fact of reworking the whole thing to include the HM triggers and D_s seemed to me to justify putting it as a new macro, but I have no particularly strong feelings on naming conventions etc. We can of course consider how to consolidate these at some point.
Cheers,
--Jeremy

miweberSMI:
Hi @jezwilkinson,

thanks a lot, I am merging now. The discussion on naming can also be postponed to later (probably a general DPG cleaning action with consistent naming of all macros would be nice).

Cheers,
Michael

PR #296, opened 2019-08-10
Flag gen headers with suppressed particles

shahor02:
@miweberSMI This will flag the headers of bg. event by "particles_suppressed" in the title, still keeping the NProduced untouched. Such headers will not be processed in the AliMCEvent::AssignGeneratorIndex, but this will require https://github.com/alisw/AliRoot/pull/992
(from compilation point of view these two PRs are independent).


miweberSMI:
Dear all, 

I am merging this PR now... Thanks a lot @shahor02 for preparing it!

Cheers,
Michael

PR #297, opened 2019-08-12
First version of LMee HIJING + injected virtual photons

miweberSMI:


miweberSMI:
Dear all, 
this PR is not for a specific production, but as preparation for a possible production for LMee (ML for HF separation).
Cheers,
Michael

PR #305, opened 2019-09-10
cd to output galice file before recreating TE tree

shahor02:


miweberSMI:
Hello,

I just tested locally 50 events in one HIJING event (same as JDL) and it works (no error messages in extractembedded.log and QA and AOD stage run smoothly). Super thanks a lot, @shahor02!

Cheers,
Michael

PR #307, opened 2019-09-11
Support special nuclei reconstruction

mpuccio:
This is to support the special PID for tracking for nuclei being introduced with alisw/AliRoot#1017
Needs to be tested before merging!

cc. @akalweit

### DIFF
--- a/DataProc/PPass/main_rec.C
+++ b/DataProc/PPass/main_rec.C
@@ -129,6 +129,13 @@ void main_rec(const char *filename="raw.root", const char* options="")
   rec.SetDeleteRecPoints(delRecPoints.Data()); 
   //
 
+  const char* confPID = gSystem->Getenv("CONFIG_HE3_PION_THRESHOLD");
+  if (confPID) {
+    int threshold = atoi(confPID);
+    threshold = threshold ? threshold : 130;
+    rec.SetPIDforTrackingOptimisedForNuclei(threshold);
+  }
+
   // Set 0.7% as fraction of friends (Ruben, 2017-08-22)
   rec.SetFractionFriends(0.007);
 
--- a/DataProc/PPass/main_rec.C
+++ b/DataProc/PPass/main_rec.C
@@ -129,7 +129,7 @@ void main_rec(const char *filename="raw.root", const char* options="")
   rec.SetDeleteRecPoints(delRecPoints.Data()); 
   //
 
-  const char* confPID = gSystem->Getenv("CONFIG_HE3_PION_THRESHOLD");
+  const char* confPID = gSystem->Getenv("ALIEN_JDL_CONFIG_HE3_PION_THRESHOLD");
   if (confPID) {
     int threshold = atoi(confPID);
     threshold = threshold ? threshold : 130;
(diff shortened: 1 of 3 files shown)


miweberSMI:
Hi @mpuccio, @akalweit, all,

what is the status of this PR?

Thanks and cheers,
Michael

ktf:
any updates?

mpuccio:
Ready to be merged for me!

MC/ReconstructionConfig.C
```diff
@@ -203,6 +205,16 @@ void ReconstructionConfig(AliReconstruction &rec, int tag_tmp)
     rec.SetRun1PIDforTracking(kTRUE);
     return;
 
+  case kReconstructionForNuclei:
+    ReconstructionDefault(rec);
+    const char* conf = gSystem->Getenv("CONFIG_HE3_PION_THRESHOLD");
+    int threshold = 130;
+    if (conf && !conf[0]) {
```
fprino, line 212: Hi Max, why did you put the condition !conf[0] ?

DataProc/PPass/main_rec.C
```diff
@@ -129,6 +129,13 @@ void main_rec(const char *filename="raw.root", const char* options="")
   rec.SetDeleteRecPoints(delRecPoints.Data()); 
   //
 
+  const char* confPID = gSystem->Getenv("CONFIG_HE3_PION_THRESHOLD");
```
fprino, line 132: Hi,
if I am not wrong, the environment variable that are set in the JDL (and I guess we will need to do it via JDL also for this special reco) have a prefix ALICE_JDL that is added automatically.
@catalinristea can confirm.
If this is the case, you should use ALICE_JDL_CONFIG_HE3_PION_THRESHOLD here.

MC/ReconstructionConfig.C
mpuccio: That is the old piece of code, I now replaced it with the one that uses only atoi, thanks for noticing!

DataProc/PPass/main_rec.C
mpuccio: I gather from the other variables around that the prefix is `ALIEN_JDL_`, I adapted the new variable to this naming convention. But let's see what @catalinristea says! Thanks for noticing also this.

chiarazampolli: Yes, I also confirm that you need the prefix if you want to read an env variable set in the JDL.

catalinristea:
Ciao,

I'm testing in grid to make sure it works. I'm not sure about the underscores in the name.
Thanks.

Yours,
Catalin.

DataProc/PPass/main_rec.C
catalinristea: Ciao,

I'm testing in grid to make sure it works. I'm not sure about the underscores in the name.
Thanks.

Yours,
Catalin.

chiarazampolli: Thanks Catalin. I guess you have doubts in the underscores in the "CONFIG_HE3_PION_THRESHOLD" part, right? There I don't know if they are accepted or not. We can check with @costing ?

catalinristea: Ciao,

I confirm it works. Please go on and merge, if nothing else.
Thanks!

Yours,
Catalin.

chiarazampolli: Done, thanks!

PR #309, opened 2019-09-16
In PbPb filtering MVertexer will be rerun

shahor02:
It will add to PileupVertexTracks pile-up vertices with >15 contributors.
Also the primary vertex found by the multiverterxed will be added (apart from the reconstruction
time primary vertex) tagges as a pile-up but with GetID()==-1

### DIFF
--- a/AOD/main_AODtrainRawAndMC.C
+++ b/AOD/main_AODtrainRawAndMC.C
@@ -616,6 +616,8 @@ void AddAnalysisTasks(const char *cdb_location, Bool_t isMC)
        * (R+compatibility)
        */
 
+     taskesdfilter->SetRunMVertexerForPileUp(iCollision == kPbPb ? 15 : 0);
+      
      if (isMuonOnly) {
        taskesdfilter->DisableCaloClusters();
        taskesdfilter->DisableCells();


catalinristea:
Ciao @shahor02, @chiarazampolli 

... this is a new feature in AliRoot master, would we need an AliPhysics backward compatibility protection?
Thanks!

Kind regards,
Catalin.

shahor02:
Hi @catalinristea,
No, it should be transparent for aliphysics.

Cheers
 Ruben

chiarazampolli:
Hello Ruben, 

As we saw in https://alice.its.cern.ch/jira/browse/ALIROOT-8339, and actually I noticed also yesterday in some tests that I was running, one needs to have the updated master (or next v5-09-51) to have this working. 
Francesco implemented a check in his branch, see https://github.com/fprino/AliDPG/commit/2f79ce8bebab4ee6eda1fbb5fc8e040b63ee6e94, we will merge it after validating that it works.

Cheers,

Chiara

shahor02:
Hi @chiarazampolli 
I thought @catalinristea was asking about the compatibility of aliphysics with aliroot, which should not have any problem since no method was removed or modified in the aliroot. As for the DPG macros vs aliroot compatibilty, indeed, since a new method of aliroot is called from the AODtrain, this latter will require recent aliroot version. @fprino thanks for adding protection, should have realized this before.
Cheers,
 Ruben

PR #310, opened 2019-09-17
[ALIROOT-8339] Non-prompt J/psi produced in B-hadron jets

yatowoo:
For J/psi in jets analysis, we proposed to generate non-prompt J/psi from B-hadron decay by pythia6.

* Create standalone macro to test the generator configuration.

Similar with HFjets, we can add b-bbar (as non-prompt) events by process kPyBeautyppMNRwmi. Then forced it decay (kBJpsiDiElectron) into J/psi with cuts on child (pT and Y). Obviously, StackFillOpt should be removed for both cases to keep the entire event. However, pT hard bin will be applied to enhance high pT jets samples.

The macro is from @miweberSMI.

### DIFF
--- a/MC/CustomGenerators/PWGDQ/Pythia6_Perugia2011_JpsiBJet13TeV.C
+++ b/MC/CustomGenerators/PWGDQ/Pythia6_Perugia2011_JpsiBJet13TeV.C
@@ -1,0 +1,51 @@
+AliGenerator *
+GeneratorCustom()
+{
+  // load libraries to use Evtgen
+  gSystem->Load("libPhotos");
+  //gSystem->Load("libEvtGenBase");
+  //gSystem->Load("libEvtGenModels");
+  gSystem->Load("libEvtGen");
+  gSystem->Load("libEvtGenExternal");
+  gSystem->Load("libTEvtGen");  
+  //
+  // set external decayer
+  TVirtualMCDecayer* decayer = new AliDecayerPythia();
+  decayer->SetForceDecay(kAll);
+  decayer->Init();
+  gMC->SetExternalDecayer(decayer);
+
+  //Generating a cocktail
+  AliGenCocktail *gener = new AliGenCocktail();
+  gener->UsePerEventRates();
+
+  AliGenPythia *pythiaHighPt = new AliGenPythia(-1);
+  pythiaHighPt->SetMomentumRange(0, 999999.);
+  pythiaHighPt->SetThetaRange(0., 180.);
+  pythiaHighPt->SetYRange(-2., 2.);
+  pythiaHighPt->SetPtRange(0, 1000.);
+  pythiaHighPt->SetProcess(kPyBeautyppMNRwmi);
+  pythiaHighPt->SetEnergyCMS(energyConfig);
+  pythiaHighPt->SetTune(kPythia6Tune_Perugia2011);
+  pythiaHighPt->UseNewMultipleInteractionsScenario();
+  // Cuts on child - J/psi
+  pythiaHighPt->SetCutOnChild(1);
+  pythiaHighPt->SetPdgCodeParticleforAcceptanceCut(443);
+  pythiaHighPt->SetChildYRange(-1, 1);
+  //pythiaHighPt->SetChildPtRange(9, 50.); // high pT J/psi would be enhanced with the pT hard bins?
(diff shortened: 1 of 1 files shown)


yatowoo:
Hi, @miweberSMI 
if I understood correctly with the discussion in mail list, the current macro is ready for a test run. Could you please give me some instructions or previous configurations for running? I am looking at the MC tutorail from AliDPG, but still unclear with the exact options of dpgsim.sh

Thanks,
Yitao

miweberSMI:
Hi @yatowoo,

the easiest way is probably to follow what was done in LHC19c6:

Packages = 
"VO_ALICE@AliDPG::prod-201903-05-1",
"VO_ALICE@AliPhysics::v5-09-31a-01-1"

$ALIDPG_ROOT/bin/aliroot_dpgsim.sh --run 294925 --mode full --uid 1 --nevents 1 --generator PWGDQ:XXX (use the name of this macro here)

You will need the OCDBs for that:
"LF:/alice/sim/2018/LHC18k3/OCDB/294925/OCDBsim.root",
"LF:/alice/sim/2018/LHC18k3/OCDB/294925/OCDBrec.root"
If for some reasons this does not work you can also download them yourself by using the command --mode ocdb,full instead.


Cheers,
Michael

PS: I would test it myself, but at the moment I don't have a running system to connect to CERN... If there are problems, please write a private mail to me!

tdahms:
Hi,

Sorry for being a bit late but I still wanted to comment: even if Jpsi2ee is moving to use EvtGen also for the B->J/ψ decay and not only for the J/ψ->ee part, in this particular case it's fine to keep PYTHIA for the B->J/ψ decay (which is what is in this PR), I believe. We don't really care about the details of the b-hadron decay but instead can benefit from filtering directly on the J/ψ kinematics, which is not supported with the current EvtGen interface. About the pT cuts on the J/ψ: it's true that the pt hard bins will enhance the high-pT J/ψ, nevertheless it might be convenient to apply some minimum pT cut if we're sure we will not look at low-pT J/ψ. At this point @yatowoo requires a pT of at least 5 GeV in his analysis. Do you plan on looking below or can apply such a cut at the generator level?

Best,
Torsten

miweberSMI:
Hi @tdahms, @yatowoo, all,

I just tested the PR on lxplus (only sim and rec) and it is running fine, but needs this PR:
https://github.com/alisw/AliRoot/pull/918 which I am not sure if it would be needed to be backported for the planned production. This is related to EvtGen, so if we woudl remove that part it would probably run w/o problems. I am not sure I fully understand your comment about Pythia @tdahms, since we use EvtGen here... Perhaps also @ffionda coudl comment.
On the pT cut I let @yatoo reply.

Cheers,
Michael

tdahms:
Hi @miweberSMI,

Sorry, I missed that there was a mistake. The line
`gene->SetParticleSwitchedOff(AliGenEvtGen::kHFPart);`
should have been
`gene->SetParticleSwitchedOff(AliGenEvtGen::kCharmPart);`
as it tells EvtGen to decay charmed particles, i.e. the ones we told PYTHIA not to decay with 
`pythiaHighPt->SetForceDecay(kBJpsiUndecayed);`
a few lines earlier. `kHFPart` should be used in case EvtGen should also decay all b-hadrons, i.e. in combination with
`pythiaHighPt->SetForceDecay(kNoDecayBeauty);`

So there's no need to wait for the PR [alisw/AliRoot#918](https://github.com/alisw/AliRoot/pull/918) but we'll need to fix this macro.

Best,
Torsten

yatowoo:
Hi, @miweberSMI, @tdahms, all,

Thanks a lot for the comments. For me, it would be pretty nice to focus on high-pT region (> 5 GeV). Maybe the people using TRD need the lower pT region (to 2 GeV). But this is not in my plan. I was wondering if it is fine to request such MC production only used for single analysis.

The changes will be applied soon.

Regards,
Yitao

yatowoo:
Hi, @miweberSMI ,

Sorry, I found the new commit on yatowoo:jpsiBJet can not be added into this PR. Do I have to create a new PR instead?

Regards,
Yitao

miweberSMI:
Hi @yatowoo 

yes, please open a new PR. And thanks to @tdahms for the extra insights!!!

Thanks and cheers.
Michael

PR #311, opened 2019-09-18
Removing unneeded arguments from FMD-related AddTasks

chiarazampolli:
Since the parameters are automatically determined from the ESDs,
it is better to not have them here - some we cannot get at this
point (e.g. B field), so better to not call them, even with
the default value, for simplicity.


chiarazampolli:
Dear @catalinristea , @fprino , @cholmcc

This is the change for the FMD tasks. For both I don't use the parameters that are automatically determined from ESDs, but only the flag is/isnot MC.

If you agree, we can merge it.

Chiara

fprino:
Hi Chiara,
thanks for implementing the change. For me it is fine to merge it.
Ciao, Francesco

catalinristea:
Ciao Chiara,

Perfect, thanks!

Yours,
Catalin.

chiarazampolli:
Thanks guys! I will merge now.

Chiara

PR #313, opened 2019-09-23
Add Config for XiC production in pp collisions with Pythia6 

gluparel:
and correct BR in custom decay table


miweberSMI:
Hi @gluparel, all,

I tested the configuration and it is running fine (only tested sim and rec). So I am merging this PR...

Cheers,
Michael

PR #316, opened 2019-10-01
Redo V0s before the gamma conversion task in AOD filtering

chiarazampolli:
Dear @catalinristea , @fprino , 

This is the change to have the V0 redone at filtering time, which we discussed to have as default. Can you check that everything is fine?

Pinging also @ddobrigk and @shahor02 , for information. 

@ddobrigk , the settings are those that you mentioned in https://alice.its.cern.ch/jira/browse/ALIROOT-8272?focusedCommentId=235244&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-235244

Chiara

### DIFF
--- a/AOD/AODtrainRawAndMC.C
+++ b/AOD/AODtrainRawAndMC.C
@@ -18,6 +18,7 @@ void AODtrainRawAndMC(Int_t merge=0, Bool_t isMC=kFALSE, Bool_t refilteringMode=
   gROOT->LoadMacro("$ALICE_ROOT/ANALYSIS/ESDfilter/macros/AddTaskESDFilter.C");
   gROOT->LoadMacro("$ALICE_PHYSICS/PWGHF/vertexingHF/macros/AddTaskVertexingHF.C");
   gROOT->LoadMacro("$ALICE_PHYSICS/PWGDQ/dielectron/macros/AddTaskJPSIFilter.C");
+  gROOT->LoadMacro("$ALICE_PHYSICS/PWGLF/STRANGENESS/Cascades/Run2/macros/AddTaskWeakDecayVertexer.C");
   /*
   gROOT->LoadMacro("$ALICE_PHYSICS/PWGJE/macros/AddTaskJets.C");
   gROOT->LoadMacro("$ALICE_PHYSICS/PWGJE/macros/AddTaskJetCluster.C");
--- a/AOD/main_AODtrainRawAndMC.C
+++ b/AOD/main_AODtrainRawAndMC.C
@@ -525,6 +525,24 @@ void AddAnalysisTasks(const char *cdb_location, Bool_t isMC)
     AddTaskCentralMult(useMC);
   }
 
+  // redo V0s
+  AliAnalysisTaskWeakDecayVertexer *taskWDV = AddTaskWeakDecayVertexer();
+  taskWDV -> SetUseImprovedFinding();
+  //V0-Related topological selections
+  taskWDV -> SetV0VertexerDCAFirstToPV(0.05);
+  taskWDV -> SetV0VertexerDCASecondtoPV(0.05);
+  taskWDV -> SetV0VertexerDCAV0Daughters(1.50);
+  taskWDV -> SetV0VertexerCosinePA(0.95);
+  taskWDV -> SetV0VertexerMinRadius(0.9);
+  taskWDV -> SetV0VertexerMaxRadius(200);
+  //Cascade-Related topological selections
+  taskWDV -> SetCascVertexerMinV0ImpactParameter(0.05);
+  taskWDV -> SetCascVertexerV0MassWindow(0.008);
+  taskWDV -> SetCascVertexerDCABachToPV(0.05);
+  taskWDV -> SetCascVertexerDCACascadeDaughters(2.0);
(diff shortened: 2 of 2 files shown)


fprino:
Hi Chiara,
thanks for taking care!
The modification looks fine to me. It would be great to get a confirmation from David about the cut values used in the task configuration.

Ciao, Francesco

chiarazampolli:
Note also that if we merge, this will become default also for PbPb, which might need different cuts. We can implement an option. (or maybe we want to enable this only in pp for now?)
Chiara

fprino:
Hi,
we can add an flag to control the re-doing of the V0s.
This flag we can set to true/false based on the collision system for the moment, or maybe based on an parameter that can be configured from the JDL?

Fot eh cuts: once we have a validated set of cuts for Pb-Pb, we can set them in the macro based on the collision system.

Ciao, Francesco

chiarazampolli:
Ciao Francesco,

I referred to the collision system indeed, when I wrote that we can implement an option, I meant that the needed information are already there. 
Let's first hear @ddobrigk 's opinion.

Chiara

ddobrigk:
Ciao Everyone, 

I confirm that this looks okay, but I also confirm that this should not be used in Pb-Pb. For that, we'd have to implement a revised filtering method (which we discussed in late July but it never fully converged into something that was validated - not obvious if we can manage to validate that before QM given the number of things that are happening now...)

Thank you! 
Cheers,
David

chiarazampolli:
Ciao David,

Great! Then I will enable this only for pp.

Cheers,

Chiara

chiarazampolli:
Ok, the changes that we discussed are now there. Can you check? then we can merge (if all ok).

Chiara

FriederikeBock:
Hi Chiara,
I would guess you can enable it for pp & pPb. 
Regards,
Friederike

chiarazampolli:
ciao @FriederikeBock , 

Fine with me, let's hear from @ddobrigk first, though, for safety.

Chiara

ddobrigk:
Ciao @chiarazampolli, 

I confirm this is also fine for p-Pb (at any energy), thanks! 

Cheers,
David

chiarazampolli:
Dear @ddobrigk , 
do you have comments about enabling the task in pPb too? 
Also note that in https://alice.its.cern.ch/jira/browse/ALIROOT-8318?focusedCommentId=235662&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-235662, they asked for different settings. So we were thinking whether to have as default ones those that we have now, and maybe implementing a switch for special cases. This is error prone, true, so better ideas are welcome.
Cheers,
Chiara

ddobrigk:
Dear @chiarazampolli, 

As I mentioned in my comment from a week ago, this should indeed be fine for p-Pb as well. Regarding the other settings: in the past, I created some auto-setup methods for the weak decay vertexer (no arguments needed) like: 

    void SetupStandardVertexing();
    void SetupLooseVertexing();

One way to proceed would be to add extra "Setup[xxx]();" functions to the weak decay finder. This isn't yet ideal, but at least it encapsulates the config calls and versions them. What do you think? 

Thank you and best regards,
David

chiarazampolli:
Ciao @ddobrigk ,

Sorry that I missed your comments, I added mine without checking above it :-)
The SetupStandardVertexing() would replace the lines:

> taskWDV -> SetUseImprovedFinding();
    //V0-Related topological selections
    taskWDV -> SetV0VertexerDCAFirstToPV(0.05);
    taskWDV -> SetV0VertexerDCASecondtoPV(0.05);
    taskWDV -> SetV0VertexerDCAV0Daughters(1.50);
    taskWDV -> SetV0VertexerCosinePA(0.95);
    taskWDV -> SetV0VertexerMinRadius(0.9);
    taskWDV -> SetV0VertexerMaxRadius(200);
    //Cascade-Related topological selections
    taskWDV -> SetCascVertexerMinV0ImpactParameter(0.05);
    taskWDV -> SetCascVertexerV0MassWindow(0.008);
    taskWDV -> SetCascVertexerDCABachToPV(0.05);
    taskWDV -> SetCascVertexerDCACascadeDaughters(2.0);
    taskWDV -> SetCascVertexerCascadeMinRadius(.5);
    taskWDV -> SetCascVertexerCascadeCosinePA(.95); 

> 
right? This would be great. Also to add some more of these. But in the AOD macro, we will still need a flag, to decide which one to call.

Chiara

fprino:
Hi,

adding methods with predefined configuration in AliPhysics is of course a good way to go.
We need however a switch in the AOD-filtering macro, as pointed out by Chiara, to select the cuts. And we need also to take care of adding some protections in this macro in AliDPG because it could be run also with older AliPhysics versions in which the methods to setup the cuts are not there.
I think it is doable: there are root methods to check if a given method/function exists in side a class, we already use them in the DPG macros.

Best regards, Francesco

chiarazampolli:
Hello,

Now the task is enabled also in pPb. I would for now merge with what we have, so that it is enabled, and then continue on the optimization of the configuration with further changes. 

Is it ok?

Chiara

fprino:
Looks OK to me.
Thanks! Ciao, Francesco

chiarazampolli:
Thanks, merging.
Chiara

PR #320, opened 2019-11-01
Protection for RAWPRODTYPE variable in case of merging

fprino:


fprino:
Hi @catalinristea 
this protection should fix the issue you reported for the merging of AOD files in case of refiltering productions.
Can you please have a look to see that I protected the correct variable?

Thanks!

Cheers, Francesco

catalinristea:
Ciao Francesco,

Looks good, thanks a lot - we cannot technically test it before the new tag, but it will certainly fix the re-filtered AODs merging. With the new tag, we'll test it, and if there are any issues we'll fix them at that point. But this shouldn't be the case.

Yours,
Catalin.

PR #332, opened 2020-01-08
Add config for HF and D2H MC productions with Pythia8

fgrosa:
- Pythia8_Monash2013_HF001.C: config for HF MC productions. For the moment only the option with forced hadronic decays is implemented. 

- Pythia8_Monash2013_D2Hdedicated.C: config for MC productions with Pythia8 with trigger particles for D2H specific channels for ML studies. For the moment only two options implemented (Ds and half Ds half D+).


chiarazampolli:
Ciao @fprino , 
Do you think we can merge? I would say so. 
For reference for this PR: see also https://alice.its.cern.ch/jira/browse/ALIROOT-8371
Chiara

fprino:
Hi,

yes, I already checked the Config files on Jira and I agree to merge them.
I am going to merge now.

Ciao, Francesco

PR #334, opened 2020-01-15
Added generators for b-jets productions with EvtGen for B decays

fcolamar:


gconesab:
Hello, 
I think both macros are ok (I will test them another day), although I am not sure about the pp generator, maybe just my ignorance in such productions. 

There is already one configuration macro named PYTHIA_HFjetpp.C that is basically the same as this one including the b process, if I am not wrong. Would it not make more sense to update the existing macro including the new optional b-jet process?

Also, it is a pythia6 generator but there is a call to pythia8 environment, is this really needed?

Regards ... Gustavo

fcolamar:
Hello Gustavo,
about updating of the existing macro: yes, this is indeed also possible, for how I modified the macro it will also be backward-compatible (in case e.g. one wants to reproduce past samples run with the original macro). This would hold also for p-Pb.
If you prefer we can proceed in this way instead of creating new files.

About the loading of the pythia8 enviroment, instead, this is something that I found out by looking at other generators containing EvtGen (the addition of the libraries is also adviced in the EvtGen instructions at https://github.com/alisw/AliRoot/tree/master/TEvtGen). At first I though it was just an example, but testing the macros resulted in execution errors in case I commented out the pythia8 environment, so I had to add it.
Unfortunately I'm not expert on this particular topic, so I have no precise information of why this seems to be required.

Dear regards,
Fabio

fcolamar:
Ciao Gustavo, all,
I have made another commit (also included in the current pull request) to simply update the two existing generator files, instead of creating new files.
Hope it is ok now.
Dear regards,
Fabio

gconesab:
Ciao Fabio,
thanks, the changes look fine, just for my knowledge:

- for what is the pythia8 environment setting and 
- why is this last generator added,  I am not sure I understand what this does? Would this not conflict when just the c quark generators are requested?

Regards ... Gustavo

fcolamar:
Ciao Gustavo,
about your two questions:
- for the first, pythia8, see the second part of my first reply
- for the second. About its purpose: it just activates the EvtGen package for having the beauty hadrons decayed instead of PYTHIA (for which the b-decays are correspondingly disabled via  kPythia6HeavyDecay_All_bDecaysEvtGen case in GeneratorPythia6Heavy, when idecay==4 is called). You are actually totally right about the conflicting, and thanks for noticing: this was fine when we intended to upload a new generator file, to be used only with EvtGen decays, but now that we just modified the already existing macros to just add another case, the new generator call needs to stay in an if(idecay==4) statement. I modified the two macros accordingly and fixed with a new commit.
Cheers,
Fabio

fcolamar:
Hello all,
is there any update on this pull request?
I'm asking just because a new tag including the updated generator files is needed to run a MC production needed for an analysis aiming HP, so it would be great to have it available in a short amount of time.
Thanks, dear regards,
Fabio

gconesab:
+1
Let's go ahead, I wanted to test it but I did not find the time

PR #335, opened 2020-01-20
add feeddown XiC and OmegaC

gluparel:
Dear all,
I would like to pull these modifications in the master. The new config and decay table update the previous files and allow the simulation of non-prompt XiC, and prompt and non-prompt omegaC0.
Thanks a lot!
Grazia


gconesab:
Ciao Grazia,

I do not see any problem in the commit, anyway, could it be possible to redo the commit with the same formatting as before so that the diff gives the real differences between the current version and the previous one, specially on the data table?

Also, I would like to test it, a simple sanity check that it does not crash, could you give me an example of the appropriate list of parameters  
$ALIDPG_ROOT/bin/aliroot_dpgsim.sh --run 286936 --uid 1 --mode sim,ocdb --nevents 1 --generator Custom --system pp --energy 13000

Grazie ... Gustavo

gluparel:
Dear Gustavo,
I have tried to do as you suggested and I hope that I haven't done too many errors in reverting/pushing... in case let me know and I will try to solve.
In the latest commit the changes with respect the previous version.

For testing you can use:
$ALIDPG_ROOT/bin/aliroot_dpgsim.sh --run 252371 --mode full --uid 1 --nevents 1 --generator Custom --reconstruction NoSDD --ocdb alien

Cheers
Grazia

gconesab:
+1
Thanks Grazia, now the diff is good. I think all is fine. I managed to test and saw Chi particles with few GeV generated in few events so I think we can approve.

PR #336, opened 2020-01-24
Threshold moved to 65

cterrevo:


gconesab:
Ciao Cristina
Is this change needed for a coming production request?
Since tracking this value changes could be difficult, should you not pass this value as an argument somehow?
Thanks

cterrevo:
Ciao, 
yes, this is needed for a HF MC production request. We usually use this macro as it is for the high multiplicity productions. Probably we can write in the JIRA ticket what is the multiplicity threshold that we use. I don't know if changing the macro adding a parameter, that then it has to be included in the command line by Catalin,  for the production, is a clean solution. I should contact Catalin and Francesco and propose the change, in case.

PR #346, opened 2020-02-27
Add a custom generator for hypertriton

fmazzasc:


mpuccio:
Ciao @fmazzasc,
can you put as pt limit 0?
Thanks a lot!
Max

PR #358, opened 2020-04-08
Script & macro to filter the raw data before processing (ALIROOT-8432)

chiarazampolli:


chiarazampolli:
Hello @catalinristea , @fprino , @mpuccio , Alex (I don't know Alex's git account, if you do, can you add him?), 

Here the commit to take the ESDs from alien, and filter only the interesting events. Please, have a look. I tried it locally, and it works.

mpuccio:
@akalweit

catalinristea:
Ciao Chiara,

Looks good, but just in case, wouldn't we add the possibility to have more than just one event from a given chunk? I would skip "found" (doesn't matter if we read all the file for every chunk, it will be reasonably fast), add a list with the events found, which then has outside the "while" loop a separate iteration to actually add the events found to fLocalList.

Thanks a lot!

Yours,
Catalin.

chiarazampolli:
Ciao @catalinristea .

I used the formatting of the file, where the same chunk is listed multiple times when needed one after the other, then other chunks appear. So, it is indeed the case that several events are taken for the same chunk, but only if they are one after the other. Do you think it is a limitation? I can remove the "found", its aim was just to not go through the file for nothing.

Chiara

catalinristea:
Ciao Chiara,

Probably we'll be fine like you said, so it would be enough to just extend to consecutive lines.
Thanks!

Yours,
Catalin.

DataProc/Common/cloneraw.C
```diff
@@ -0,0 +1,79 @@
...
+      TObjArray* arr = strTmp.Tokenize(":");
+      TObjString* objstr = (TObjString*)arr->At(2);
+      //cout << "token 2: " << objstr->String() << endl;
+      Int_t nevToFilter = (objstr->String()).Atoi();
+      fLocalList.Enter(nevToFilter);
+      cout << "Found event " << nevToFilter << " for chunk " << chunkName << endl; 
+      found = true;
+      delete arr;
+    }
+    else if (found == true) break;
```
fprino, line 51: Isn't this condition preventing from finding the 2nd line/event in those cases in which there are two events from the same chunk?

fprino:
HI,

I have a doubt on the usage of the found. I wrote a comment in the code.
But it nay be that I overlooked some parts of the logic to parse the file...

Ciao, Francesco

chiarazampolli:
Ciao, 

It works both for 1 event and multiple, see the following examples:

1 event:

`root [0] .x cloneraw.C("18000296549030.2810.root", "18000296549030.2810")
=> Trying to connect to Server [1] root://pcapiserv08.cern.ch:10000 as User zampolli 
W-TEnvRec::ChangeValue: duplicate entry <Library.vector<short>=vector.dll vectorbool.dll> for level 0; ignored
Opening ESD file alien:///alice/data/2018/LHC18q/000296549/pass3/18000296549030.2810/AliESDs.root
I-TAlienFile::Open: Accessing image 1 of alien:///alice/data/2018/LHC18q/000296549/pass3/18000296549030.2810/AliESDs.root in SE <ALICE::FZK::SE>
W-TClass::TClass: no dictionary for class pair<pair<int,int>,int> is available
Found event 9 for chunk 18000296549030.2810
`

2 events:
`root [1] .x cloneraw.C("18000296549028.1208.root", "18000296549028.1208")
=> Trying to connect to Server [0] root://pcapiserv03.cern.ch:10000 as User zampolli 
Opening ESD file alien:///alice/data/2018/LHC18q/000296549/pass3/18000296549028.1208/AliESDs.root
I-TAlienFile::Open: Accessing image 1 of alien:///alice/data/2018/LHC18q/000296549/pass3/18000296549028.1208/AliESDs.root in SE <ALICE::FZK::SE>
Found event 6 for chunk 18000296549028.1208
Found event 102 for chunk 18000296549028.1208
`

For me all is fine.

Chiara

DataProc/Common/cloneraw.C
chiarazampolli: Ciao, 

The break is called when we don't find anymore the wanted chunk in the line, so when there are multiple events, it is not yet called, unless the chunks are not in order. @mpuccio , @akalweit, can you confirm that when a chunk has more than one event, they are all consecutives? In the file, the order is not alphabetical. We can also change the file, and make it alphabetical, to be sure. Or we loop over all lines. 

Let me know what you prefer.

Chiara

chiarazampolli: The new commit takes as input files that are sorted alphabetically - there is no real change in the logic, just the file name. So, beforehand, we should sort the file. 
But as I said, if we want to be 100% sure, we can go through all the lines every time.

akalweit: Hi,
I do not have a strong opinion on that. For me, it would also be fine to sort the file beforehand.

Thanks a lot,
Alexander

akalweit:
Ciao Chiara,
thanks a lot. I browser your macros and scripts and it looks okay to me at first glance. As soon as there are say 100 events available somewhere, I can also run over them for detailed inspection.
Cheers,
Alexander

chiarazampolli:
Ciao Alex, 

Why 100 events? The macro will create 1 filered raw per chunk that you provided, so we will have 1-2 events per new chunk. This because the filtering will become part of the processing. 
We can also save the output raw data. 

Chiara

DataProc/Common/cloneraw.C
```diff
@@ -0,0 +1,79 @@
...
+  // --- End of "Analysis part"
+
+  // --- "Clone part"
+
+  // Simple example how we can clone the tree. Of course in rallity we have to use
+  // chain of RAW files that corresponds exactly to the chain of ESDs. Since the GUID of
+  // the RAW file is not stored in the ESD tag file, we will have to add it.
+
+  // RAW file and tree
+  TFile fraw("fileName");
```
catalinristea, line 67: Shouldn't this be?
  TFile fraw(fileName);

DataProc/Common/cloneraw.C
```diff
@@ -0,0 +1,79 @@
...
+
+  // from env variables
+  Int_t runNumber = atoi(gSystem->Getenv("ALIEN_JDL_LPMRUNNUMBER"));
+  Int_t year = atoi(gSystem->Getenv("ALIEN_JDL_LPMANCHORYEAR"));
+  TString period = gSystem->Getenv("ALIEN_JDL_LPMPRODUCTIONTAG");
+  TString pass = gSystem->Getenv("ALIEN_JDL_LPMPASSNAME");
+  
+  // ESD file and tree
+  TGrid::Connect("alien://");
+  TString filename(TString::Format("alien:///alice/data/%d/%s/%09d/%s/%s/AliESDs.root", year, period.Data(), runNumber, pass.Data(), chunkName));
```
catalinristea, line 14: ALIEN_JDL_LPMPASSNAME is the output path, where we don't have ESDs for event id matching. I have added into the JDL a new variable, called (ALIEN_JDL_)ESDREFNAME which can be used to point to the paired ESD location.

DataProc/Common/cloneraw.C
chiarazampolli: Ciao Catalin,

Of course! Sorry, I edited the file directly here, and missed the inverted commas... 
Thanks for spotting!

Chiara

DataProc/Common/cloneraw.C
chiarazampolli: Ciao Catalin, 

Isn't this the pass name, e.g. "LPMPassName=pass3" from https://alimonitor.cern.ch/jobs/jdl.jsp?pid=1822219287? I just need the pass name... It is also ok to add another JDL variable, of course, but if we can build this from what we have, isn't it easier?

Chiara

catalinristea: Ciao Chiara,

For a new production, LPMPASSNAME is the output folder, ie "filter_pass3" from JDL /alice/data/2018/LHC18r/$1/filter_pass3/#alienfilename/.root//#. This is the test production, so input ESDs are not there and it fails. If we use the same location as pass3, then we'll put the output into the same folder as standard production, and this may work only if we output different files, since original Pass3 would have already ESDs. Also, as for this new production, we run full reco, resulting in new ESDs, we'll clash between original ESDs and this new one. Even if we stop and produce just raw filtered raw1.root file, it won't be wise to mix 2 different productions (standard Pass3 and this new type), because removing one production from MonALISA will result in the entire output lost. Or do you think to run such filtering during standard pass?

Thanks!

Yours,
Catalin.

catalinristea: ...this is why we need a new variable to set the location from where to get input ESDs. We cannot use LPMPASSNAME, as this is the output path of the (new) production.
Thanks!

Yours,
Catalin.

chiarazampolli: Ciao Catalin, 

Now I understand, I did not know, sorry. Then I guess we need a new JDL tag, I cannot think of a different way...
For now, I don't plan such filtering at "pass" time, it could happen (e.g. for 15o pass2). Note that it can happen only in case of reprocessings, because only after a first pass one hase the list of interesting events. But anyway, also during a pass, if we have the extra ta, we will be fine. 
Now I will change the code according to the new tag.

Chiara

chiarazampolli:
Ciao @catalinristea ,

Can you check that now everything is ok?

Thanks!

Chiara

catalinristea:
Ciao Chiara,

Thank you, it looks perfect.
From my side this development can be merged immediately.

Yours,
Catalin.

chiarazampolli:
Thanks @catalinristea and also @fprino for the feedback!

Chiara

PR #362, opened 2020-04-27
Add config for D2H PbPb MC with HIJING + Pythia8

fgrosa:
- Add config for D2H PbPb MC productions with hijing + pythia8 @cterrevo @jezwilkinson 

- @fprino @chiarazampolli I tested it and it seems to work (I simulated ~10 events since it is very slow, similarly to the one with pythia6), but I would like to ask you to check, in particular for the embedding and Lc options which I basically copied from [CustomGenerators/PWGHF/Hijing_HF001](https://github.com/alisw/AliDPG/blob/master/MC/CustomGenerators/PWGHF/Hijing_HF001.C). 
For the Lc in particular there are three options (two for the Lc to V0bachelor, one with pK0s only), that I implemented, by defining the trigger particle

   https://github.com/fgrosa/AliDPG/blob/5df0e30f6dccd54c402f8fd3d6f183c46c735435/MC/CustomGenerators/PWGHF/Hijing_Pythia8_Monash2013_HF001.C#L44

  and modifying the decays to be kept

   https://github.com/fgrosa/AliDPG/blob/5df0e30f6dccd54c402f8fd3d6f183c46c735435/MC/CustomGenerators/PWGHF/Hijing_Pythia8_Monash2013_HF001.C#L109

   which are all enabled only in case of channelOption == 0 (corresponding to the D2H general purpose MC for efficiencies).

### DIFF
--- a/MC/CustomGenerators/PWGHF/Hijing_Pythia8_Monash2013_HF001.C
+++ b/MC/CustomGenerators/PWGHF/Hijing_Pythia8_Monash2013_HF001.C
@@ -122,7 +122,7 @@ AliGenerator *GeneratorCustom(TString opt = "")
       (AliPythia8::Instance())->ReadString("4122:onIfMatch = 3122 211");
       (AliPythia8::Instance())->ReadString("4122:onIfMatch = 2212 311");
     }
-    if(channelOption == 2) { //general purpose or pK0s triggered
+    if(channelOption == 2) { //pK0s triggered
       (AliPythia8::Instance())->ReadString("4122:onIfMatch = 2212 311");
     }
 
@@ -137,6 +137,10 @@ AliGenerator *GeneratorCustom(TString opt = "")
     (AliPythia8::Instance())->ReadString("4132:onMode = off");
     (AliPythia8::Instance())->ReadString("4132:onIfMatch = 3312 211");
   }
+
+  // Set up2date lifetimes for hadrons
+  // lambda_b from PDG 2019: tau0 = 1.471 ps = 441 m/c = 0.441 mm/c
+  (AliPythia8::Instance())->ReadString("5122:tau0 = 4.41000e-01");
 
   TFormula *formula = new TFormula("Signals","max(1.,120.*(x<5.)+80.*(1.-x/20.)*(x>5.)*(x<11.)+240.*(1.-x/13.)*(x>11.))");
   if(isEmbedding){
--- a/MC/CustomGenerators/PWGHF/Hijing_Pythia8_Monash2013_HF001.C
+++ b/MC/CustomGenerators/PWGHF/Hijing_Pythia8_Monash2013_HF001.C
@@ -1,9 +1,9 @@
 AliGenerator *GeneratorCustom(TString opt = "")
 {
 
-  TString optList[5] = {"had", "LcK0Sp", "LcK0SpBDTsig", "LcpKpi", "DsKKpi"};
+  TString optList[4] = {"had", "LcK0Sp", "LcpKpi", "DsKKpi"};
   Int_t channelOption = 0;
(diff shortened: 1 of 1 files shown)


chiarazampolli:
Ciao @fgrosa , 

Sorry, I overlooked this PR. It is still needed, right?

Chiara

fprino:
Hi,

I also had a look at the code (sorry for the delay, but I did not get the meail notification of the PR and thanks to Chiara for writing about it in an email).
The macro looks OK to me. I just have two minor points:
- maybe it is better to set the up2date Xic0 lifetime. I know that likely we won't simulate Xic0 in Pb-Pb, but in case we do... we don't risk to forget it. This can be done via:
  (AliPythia8::Instance())->ReadString("5122:tau0 = 4.41000e-01");
- for the Lc decays in the Hijing+Pythia6 configuration macro we were forcing the decay modes to kLcpKpi and kLcpK0S for the corresponding enrichment cases, in order to increase the statistics. I never tested that these decay options work for the Pythia8 decayer, but maybe it is worth to consider them, test them and in case include them in the configuration. What do you think?

Ciao, Francesco

fgrosa:
Ciao @fprino, @chiarazampolli,
no problem, there was no hurry! Thanks for the revision, I added a couple of things that should implement the comments by Francesco:

> Hi,
> 
> I also had a look at the code (sorry for the delay, but I did not get the meail notification of the PR and thanks to Chiara for writing about it in an email).
> The macro looks OK to me. I just have two minor points:
> 
>     * maybe it is better to set the up2date Xic0 lifetime. I know that likely we won't simulate Xic0 in Pb-Pb, but in case we do... we don't risk to forget it. This can be done via:
>       (AliPythia8::Instance())->ReadString("5122:tau0 = 4.41000e-01");
Just to be sure, you meant the Lb decay length as it is in the code, is it correct? In this case I added it in this commit f58680e184798d499e34fced734bb8027fa7eed4 
> 
>     * for the Lc decays in the Hijing+Pythia6 configuration macro we were forcing the decay modes to kLcpKpi and kLcpK0S for the corresponding enrichment cases, in order to increase the statistics. I never tested that these decay options work for the Pythia8 decayer, but maybe it is worth to consider them, test them and in case include them in the configuration. What do you think?
Yes I agree! This is what I tried to implement in https://github.com/fgrosa/AliDPG/blob/6455979603360e0d8e365bee6961909185ef2d13/MC/CustomGenerators/PWGHF/Hijing_Pythia8_Monash2013_HF001.C#L111, however I relised that probably this would not be used since the config will likely use the forced decays in AliDecayerPythia8::ForceDecays. As a possible solution I forced the usage of the decays defined in the config with this commit 6455979603360e0d8e365bee6961909185ef2d13, but I am not sure whether this is the best solution.

Ciao,
Fabrizio

> 
> Ciao, Francesco

fprino:
Hi Fabrizio,

yes, the Lb lifetime, sorry for the mismatch of hadron species...
Concerning the forcing of decays: I suggest to keep using ForceHadronicD since the options to force only the pK0s and the pKpi should be available also in AliDecayerPythia8 (I never tested it, but it is implemented).

Cheers, Francesco

fgrosa:
Hi Francesco, 
thanks for the suggestion, I did not realised it before!
It should be done correctly now with this commit 66565951fa3a905ac44ac91f7d12d5f2928ded0f. Since it was not implemented in AliDecayerPythia8 I removed the options for the generic V0bachelor and I kept only the pK0s (which is also the one that we used with pythia6 if I am not wrong).

Ciao,
Fabrizio

fprino:
Hi,

thanks you Fabrizio for implementing my suggestions.
For me this PR can be merged.

Ciao, Francesco

gconesab:
Let's merge then ... Gustavo

PR #371, opened 2020-05-18
Replace TAlienCollection by TGridCollection, and replace the ::Error …

gconesab:
…string on the macro name


gconesab:
I will merge it, unless there is any comment in the next hours.

PR #395, opened 2020-07-15
Extract Calo data

shahor02:
Copy all CaloClusters having at least 1 signal label, discarding those which contain BG labels only,
for the saved clusters the BG labels will be set to -1 but not removed.
Matched eliminated tracks are removed from the cluster track list.

All cells (also having BG labels only) contributing to saved clusters will be preserved

### DIFF
--- a/MC/ExtractEmbedded.C
+++ b/MC/ExtractEmbedded.C
@@ -158,6 +158,7 @@ void FilterEvent(AliESDEvent *evIn, AliESDfriend* evInF, AliESDEvent *evOut, Ali
   int ntr = evIn->GetNumberOfTracks();
   int trStatus[ntr];
   memset(trStatus,-1,sizeof(int)*ntr);
+  TBits caloCellsID[2]; // EMC and PHOS cells
   
   // check tracks
   for (int itr=0;itr<ntr;itr++) {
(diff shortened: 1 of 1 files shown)


shahor02:
@gconesab Could you please verify changes? Note that since the clusters contributed by at least 1 signal MC label are preserved while the background tracks are removed, it may happen that the cluster fTrackDx,fTrackDz info (distance to the closest matched track) might correspond to removed track (can be checked by 1st label being -1).

I've added Print method https://github.com/alisw/AliRoot/pull/1183 to Calo Clusters and Cells ESD classes and dumped the 
clusters/cells for single event with embedding before and after running the macro:
[calout.tar.gz](https://github.com/alisw/AliDPG/files/4926447/calout.tar.gz)
````
cl_orig.txt  # original clusters
cl_emb.txt # extracted clusters
ecell_orig.txt # original emcal cells
ecell_emb.txt # extracted emcal cells
pcell_orig.txt # phos ...
pcell_emb.txt # phos ...
````

File: MC/ExtractEmbedded.C [hunk 1]
```diff
@@ -265,6 +257,98 @@ void FilterEvent(AliESDEvent *evIn, AliESDfriend* evInF, AliESDEvent *evOut, Ali
...
+  int caloClusStatus[nclIn];
+  memset(caloClusStatus,-1,sizeof(int)*nclIn);
+  for (Int_t ic=0; ic<nclIn; ic++) {
+    AliESDCaloCluster *cluster = evIn->GetCaloCluster(ic);
+    int nlab = cluster->GetNLabels();
+    TArrayI* labels = cluster->GetLabelsArray();
+    for (int il=0;il<nlab;il++) {
+      if (KeepTrack( (*labels)[il]) ) {
+        caloClusStatus[ic] = -10; // keep
+        break; // keep clusters with at least single signal contribution
```
gconesab, line 273: Is the break needed? It could be that the loop is not ended and there are still MC labels from the background not changed to -1?

shahor02: @gconesab indeed, the break was there just to avoid unnecessary checks once it is clear that the cluster should be kept, but it may leave some bg. labels unchanged. Fixed.

gconesab:
Thanks Ruben, I think it is good. I will merge.
Cheers ... Gustavo

PR #397, opened 2020-07-16
update centrality dep. injection based on bmin / bmax

ffionda:
Custom generator macro changed in order to consider a centrality dependent injection based on bmin / bmax (needed for https://alice.its.cern.ch/jira/browse/ALIROOT-8511).

### DIFF
--- a/MC/CustomGenerators/PWGDQ/Hijing_Jpsiee002.C
+++ b/MC/CustomGenerators/PWGDQ/Hijing_Jpsiee002.C
@@ -1,5 +1,5 @@
 AliGenerator *
-GeneratorCustom(TString ptParamPrompt="UserParam_PbPb5TeV_0010")
+GeneratorCustom(TString ptParamPrompt="UserParam_PbPb5TeV_0010", Bool_t useEventbyEventCent = kFALSE)
 {
   TString simulation = gSystem->Getenv("CONFIG_SIMULATION");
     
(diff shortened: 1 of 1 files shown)


miweberSMI:
Hi @ffionda 

thanks for this commit. I am wondering one thing: why is the number of injected JPsis still calculated with the same formula? I thought that you wanted to inject different numbers for each of the centrality bins:0-10,10-30,30-50,50-XX

Cheers,
Michael

ffionda:
Hi @miweberSMI 
the numbers we decided to inject per each centrality interval (i.e. 10 for 0-10%, 7 for 10-30%, 5 for 30-50%, 2 for 50-XX%) are evaluated directly from the formula (which is what we used for pass1 production). In particular, we are taking the values of injected J/psi retrieved by the formula at the middle value in the [bmin,bmax] interval. Since the option useEventbyEventCent is kFALSE we are not passing TFormula in AddGenerator() (this infact becomes 0x0), but directly the number of signals we want to inject (i.e. "ntimes"). In case the option useEventbyEventCent is kTRUE we obtain the same configuration of pass1 (i.e. we use event by event b).  I hope I understood correctly the suggestion from @fprino (I followed as example what is done in https://github.com/alisw/AliDPG/blob/master/MC/CustomGenerators/PWGHF/Hijing_HF001.C), maybe @fprino  could comment in case I'm wrong.
Thank you
Cheers

Fiorella

miweberSMI:
hi @ffionda ,

ah... my mistake. When I was plotting the formula I though of percentiles, but of course this is b (I absolutely should know this ;( ).
Then all is good from my side! Sorry for the discussion in the first place...

Cheers,
Michael

gconesab:
Since a consensus is reached I will approved.
Let me know if I should test the macro.
Gustavo

PR #398, opened 2020-07-16
adapt GA macros to embedding and pile-up configuration

gconesab:


MC/CustomGenerators/PWGGA/Hijing_Gamma001.C
```diff
@@ -3,15 +3,40 @@ GeneratorCustom()
...
+  {
+    ntimes=neutralsF->Eval(bEvalFormula);
+    delete neutralsF; neutralsF=0x0;
+  }
   AliGenerator   *pi0  = GeneratorInjector(1, 111, 0., 50., -1.2, 1.2);
   AliGenerator   *eta  = GeneratorInjector(1, 221, 0., 50., -1.2, 1.2);
-  ctl->AddGenerator(pi0,  "Injector (pi0)", 1., neutralsF);
-  ctl->AddGenerator(eta,  "Injector (eta)", 1., neutralsF);
+  ctl->AddGenerator(pi0,  "Injector (pi0)", 1., neutralsF,ntimes);
+  ctl->AddGenerator(eta,  "Injector (eta)", 1., neutralsF,ntimes);
```
FriederikeBock, line 39: I might be missing something, but isn't the AddGenerator already added ntimes if you give it a TFormula?
I think if we need the number to be fixed beforehand we need to have :
if ( isEmbedding ){
  ctl->AddGenerator(pi0,  "Injector (pi0)", ntimes);
  ctl->AddGenerator(eta,  "Injector (eta)", ntimes);
} else {
  ctl->AddGenerator(pi0,  "Injector (pi0)", 1., neutralsF);
  ctl->AddGenerator(eta,  "Injector (eta)", 1., neutralsF);
}

gconesab: Unfortunatelly this does not work in the way the pile-up is implemented. We have to pass the number of times before the event generation and thus only the average of the cycle bmin to bmax is applied to the formula once.

If the b per event is needed, a modification not existing would be needed with the corresponding porting and delays. You should decide if you can live with this or need a development that would take some time.

gconesab: I misread your comment. I just copied what was done in other macros now in production.
Looking to AliGenCocktailEntry, and the data members fRate and fNTimes comments, if I understood it correctly the difference with the current approach and what you propose is that in the current approach the Box generator is called n times and in what you propose n particles are produced in a single call to the Box generator. I doubt there is a significant difference. The current approach just has less lines.

FriederikeBock: ok now I get it, I think the implementation is correct as done.

gconesab: Thanks, approving then ...

PR #401, opened 2020-07-27
Hijing_Rsn005 version with more Lambda(1520)

preghenella:
This PR adds a new LF custom configuration `Hijing_Rsn005_moreLambda1520`.
The configuration is for the production discussed in 
https://alice.its.cern.ch/jira/browse/ALIROOT-8525

It is an extension of the `Hijing_Rsn005` setup, with `Lambda(1520)` injections happening in 2/11 slots instead of the original 1/10.


gconesab:
Fine with me, maybe one could add a comment saying what is the difference with respect the other macro, the double lambda.

PR #412, opened 2020-09-09
Add option for Xic0 semileptonic decay

fgrosa:
This modification depends on this PR on AliRoot: https://github.com/alisw/AliRoot/pull/1213. There is however a protection to avoid crashes due to the usage of AliRoot tags without the option for the Xic0 semileptonic decay.


chiarazampolli:
Looks fine to me. @gconesab , do you agree to merge?

gconesab:
Looks good to me, merging. G

PR #421, opened 2020-09-24
Support for re-filtering in user space, both MC and RAW data reconstr…

catalinristea:
…uctions

Ciao,

Here is the commit to support running AOD re-filterings as standard user. It has been tested on both user/central re-filtering, but before merging I want also to test that AOD macro is functional for standard production (to make sure I haven't broken anything). I'll let you know when these tests are completed.

Thanks!

Yours,
Catalin.


gconesab (approved):
Hi Catalin,
It looks good, the little I can follow.
Gustavo

catalinristea:
Hi Gustavo,

Thanks a lot for having a look. I have tested it for both user/central productions and standard MC and it seems fine. I think we can merge it at your convenience.

Yours,
Catalin.

chiarazampolli:
Ciao @catalinristea , 

Merged. Was this implementation triggered by some specific request?

Thanks!

Chiara

PR #422, opened 2020-09-25
Propagate energy, number of events and seed directly from dgpsim sett…

mfasDa:
…ings to aligenmc

Use parameters for number of events, energy and seed
from DPG sim directly rather then passing them as
extra arguments via genopts. In addition log settings
in gen.log.

### DIFF
--- a/MC/aligenmc/gen_aligenmc.sh
+++ b/MC/aligenmc/gen_aligenmc.sh
@@ -8,4 +8,15 @@ fi
 
 echo "running in `pwd`, writing HepMC to $1"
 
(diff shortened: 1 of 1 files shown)


mfasDa:
@gconesab @catalinristea Please review and merge if appropriate!

With the change I propagate settings from dpgsim.sh directly to aligenmc instead of handling them as extra parameters via genopts. I think this is more transparent. Furthermore it prevents that AliSimulation and the event generator are configured with a different number of events.

Settings propagated:
- Number of events
- CMS energy
- Random seed
- min. pt. hard
- max. pt. hard

gconesab:
Thanks Markus,
I do not know yet how works the aligenmc package, but I trust this is properly done.

mfasDa:
https://github.com/alisw/aligenmc/pull/5

With this pull request min. and max pt-hard are added as additional parameters. For the moment only the Herwig configuration makes use of it, but others could as well. This will need a new tag of aligenmc.

I have checked that the parameters are properly propagated.

PR #429, opened 2020-10-09
Custom generator trigger passed via dpgsim

gconesab:
As requested in DPG meeting 05/10/20, the possibility to recover the macro with event generator external trigger in the same way as for generators is implemented.
Example use cases:

- $ALIDPG_ROOT/bin/aliroot_dpgsim.sh --run 296196 --mode sim --simulation GeneratorOnly --uid 1 --nevents 10  --generator Pythia6 --gentrigger Custom (It needs to have locally the file GenTriggerCustom.C, before named UserTrigger.C)

- $ALIDPG_ROOT/bin/aliroot_dpgsim.sh --run 296196 --mode sim --simulation GeneratorOnly --uid 1 --nevents 10  --generator Pythia6 --gentrigger DPG:UserTriggerExample

Note that, 
* Now there is no need for the macros Pythia6_UserTrigger.C and EPOS_UserTrigger.C, the generator configuration is handled in the usual way. Maybe we should implement a warning for the generators this is not working, all but Pythia6 and those using AliGenExtFile
* The UserTriggerFunction must be implemented in a macro with method name GenTriggerCustom and for the "Custom" option the macro must be named GenTriggerCustom.C
* Right now, I do not see a way to pass settings to the trigger user function, which would be interesting to avoid many macros doing the same thing but just changing a number
* The PWG macros are now expected in the same directories as for the generators, it can be easily changed to other location, but I preferred to not add now more directories.

Tested on Root6 not yet on Root5, but it should.

### DIFF
--- a/MC/CustomGenerators/DPG/UserTriggerExample.C
+++ b/MC/CustomGenerators/DPG/UserTriggerExample.C
@@ -19,7 +19,7 @@
 Bool_t UserTriggerFunction(AliStack *stack)
 {
   printf("____________________________\n");
-  printf("USER TRIGGER IMPLEMENTATION \n");
+  printf("USER TRIGGER IMPLEMENTATION EXAMPLE\n");
 
   Int_t nTracks  = stack->GetNtrack();
   printf("n Tracks = %i \n",nTracks);
@@ -39,7 +39,7 @@ Bool_t UserTriggerFunction(AliStack *stack)
 //------
 /// Main, pass the generator and set the user trigger
 //------
-void UserTrigger(AliGenerator * generator)
+void GenTriggerCustom(AliGenerator * generator)
 {
   Bool_t (*funcUserTrigger)(AliStack*) = UserTriggerFunction;
   
--- a/MC/GeneratorConfig.C
+++ b/MC/GeneratorConfig.C
@@ -494,6 +494,61 @@ void GeneratorConfig(Int_t tag)
     }
   }
   
+  // Externally defined event trigger
+  
+  TString gentrigstr = gSystem->Getenv("CONFIG_GENTRIGGER");
+
+  // Custom external trigger
+  if ( gentrigstr == "Custom")
+  {
+    if ((gROOT->LoadMacro("GenTriggerCustom.C")) != 0) {
+      printf("ERROR: cannot find GenTriggerCustom.C\n");
+      abort();
+      return;
+    }
+        
+    // Load and compile macro
+    gROOT->LoadMacro("GenTriggerCustom.C+");
+    
+    gg_tmp_gen = gen;
+    gROOT->ProcessLine("GenTriggerCustom(gg_tmp_gen);");
+  }
+  else if (gentrigstr!="")
+  {
+    // PWG
+    TObjArray  *oa         = gentrigstr.Tokenize(":");
+    TObjString *pwgtrig    = (TObjString*) oa->At(0);
+    TObjString *pwgtriggen = (TObjString*) oa->At(1);
+    // In case we manage to pass an option to the function, not possible now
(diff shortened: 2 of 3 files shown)


gconesab:
Walking back home I though how to pass parameters to the external trigger via an environmental variable. For this use case one could do 
* $ALIDPG_ROOT/bin/aliroot_dpgsim.sh --run 296196 --mode sim --simulation GeneratorOnly --uid 1 --nevents 10 --generator Pythia6 --gentrigger DPG:UserTriggerExample:110

Where 110 is the parameter read by the external trigger.

catalinristea:
Ciao Gustavo,

Indeed, this last change I was to suggest to you, but you figured it out before...

Sounds perfect with this last change - maybe you want to check a bit the workflow, in order to take out what you referred before by "to avoid many macros doing the same thing but just changing a number".

Thanks a lot!

Yours,
Catalin.

gconesab:
Thanks Catalin, I was not sure I could edit the first comment, updated now.

catalinristea:
Ciao Gustavo,

Then we're just fine.
I'm not sure if the default behaviour (ie the user trigger macro in AliDPG) would be to rely on the number of particles in the stack, but rather on the number of primary charged particles within V0 acceptance (please see Markus code excerpt from AliPhysics, in the JIRA). I think the latter would be more adequate for such productions.
On this, I let Chiara and Francesco decide, from my side this pull request can be merged.

Thanks!

Yours,
Catalin.

catalinristea:
Ciao Gustavo,

One more small detail, please:
Could you please remove old generators, since the new functionality makes these unnecessary:
- AliDPG/MC/CustomGenerators/DPG/EPOS_UserTrigger.C
- AliDPG/MC/CustomGenerators/DPG/Pythia6_UserTrigger.C
Keeping these would only add ambiguity.

Thank you!

Yours,
Catalin.

gconesab:
Hi Catalin,
I was waiting for the comments to remove the Pythia and EPOS macros. Regarding the trigger macro, this one is just a simple example to show how it works. We can add the one to be used on the production and remove this one. Since you have the macro for the testing with the proper trigger you can maybe commit it after this commit, adding the recovery of the cut via the environment.

catalinristea:
Hi Gustavo,

Thanks a lot!
I certainly can update the user trigger example in this direction, if Chiara and Francesco want it. Meanwhile, tomorrow I'll have a look at ongoing tests with 56, to check we're fine.

Yours,
Catalin.

chiarazampolli:
Ciao @gconesab , 

Thanks! To me it looks fine. 
I think it does not matter much how the example UserTriggerExample is defined, it is clearly an example, one can see it in the name. But I would probably commit the macro used in production, since that seems a quite general case. Otherwise it should go under CF... 

Cannot we generalize the macro used in production with a pt cut also? For the CF request, it should do nothing, of course. Or would we then add too many parameters (pt min, pt max)?

Chiara

fprino:
Hi,

thanks Gustavo for this development.
Looks very good to me.
It is clear that the macro tat you provide is an example, so to me it looks OK for its purpose.

Ciao, Francesco

gconesab:
Ciao Chiara, Catalin, Francesco,

ok, then we can merge this as it is and in a separate commit we can put the macro for production. Since it is a rather general trigger, maybe DPG directory could be place and not PWGCF. We can generalize it with pt cuts, but then the environment variable content should be something like nMin_xx_ptMin_yy_ptMax_zz or xx_yy_zz and somehow extract xx, yy, zz on the macro.

catalinristea:
Hi Gustavo,

Sounds good to me.
Thanks!

Yours,
Catalin.

chiarazampolli:
Good to go for me too!

Chiara

PR #432, opened 2020-10-12
Use CONFIG_PROCESS directly to forward process (tune) to aligenmc

mfasDa:
With current options in aligenmc all parameters are
forwarded with dpgsim CONFIG parameters, currently
no need for aligenmc extra params (could change in
the future)


mfasDa:
@catalinristea @gconesab This is connected to https://alice.its.cern.ch/jira/browse/ALIROOT-8558

PR #433, opened 2020-10-16
add debug option to show the V0 mult user trigger prints and pass extra number of events in generator pool

gconesab:


chiarazampolli (commented):
Seems fine to me, thanks!

PR #434, opened 2020-10-20
Create Hijing_Nuclex_anti3H.C

mpuccio:


gconesab:
Hello, I guess we will need to run this with the possibility to add pile-up or reusing background so one would need to consider this possibility like in macro  Hijing_Nuclex004.C where if the configuration the string embed is found, then the hijing is produced separately.
Could you please consider it?
Thanks ... Gustavo

gconesab:
Ciao @mpuccio , will you implement the suggested changes?
Thanks ... Gustavo

gconesab:
Ciao @mpuccio , any comments or  problem to implement such modification?
Thanks ... Gustavo

mpuccio:
Hi @gconesab,
sorry for the late action and reply. Now it should be fine. @mcolocci, please check if the pt range is fine.

Cheers,
Max

gconesab:
Ciao Massimo,
you just had to do like in the macro I put as example Hijing_Nuclex004.C and wrap the hijing generator call with 

 if(!simulation.Contains("Embed")){
    AliGenerator *hij = GeneratorHijing();
    ctl->AddGenerator(hij, "Hijing", 1.);
  }

The pile-up is handled by the embedded hijing that you do not set-up in your macro.
Thanks ... Gustavo

mpuccio:
OK, updated

gconesab:
Thanks!

PR #443, opened 2020-11-10
Special configs for HF productions in Pb-Pb

fprino:
This modification in main_AODtrainRawAndMC.C allows to set automaticallythe configuration for the vertexingHF filtering in the MC simulations anchored to 2018 Pb-Pb sample.
Some notes:
- the limits on the impact parameter, which are used to "decide" which of the configuration files should be used, are set based on the ones used in LHC20g2a and LHC20g2b productions. If in the future we are going to do a production in e.g. 10-20% or 20-30% centralities we may need to come back to them.
- to define automatically the configuration file based on the centrality interval of the MC production, I used the variables CONFIG_BMIN and CONFIG_BMAX, which are defined in the simulation script (dpgsim.sh). This means that this strategy would work for the AOD creation which runs in the reconstruction job. For the AOD refilterings, we would need to set by hand these variables when preparing the refiltering. This would complicate this refilteirng jobs with a setting that is production type dependent. However, also other possibilities (such as setting the name of the configuration file in the JDL) would have the same issue in the case of refilterings.

### DIFF
--- a/AOD/main_AODtrainRawAndMC.C
+++ b/AOD/main_AODtrainRawAndMC.C
@@ -727,8 +727,18 @@ void AddAnalysisTasks(const char *cdb_location, Bool_t isMC)
     {
 
       Int_t configHF=0;
+      TString specialConfig="";
       if(iCollision == kPbPb || iCollision == kXeXe) configHF=1;
-      AliAnalysisTaskSEVertexingHF *taskvertexingHF = AddTaskVertexingHF(configHF,train_name,"",run_number,periodName);
+      if(isMC && iCollision == kPbPb && year==2018){
+	Float_t bmin=0;
+	Float_t bmax=20;
+	if (gSystem->Getenv("CONFIG_BMIN")) bmin = atof(gSystem->Getenv("CONFIG_BMIN"));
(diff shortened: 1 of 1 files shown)


catalinristea:
Ciao Francesco,

This looks good to me, with the caveat that we'll have different AOD numbers when re-filtering one given MC production with centrality dependent sub-cycles, which adds a bit of ambiguity.

For ongoing AOD242 we're then fine, but to start corresponding MC re-filterings we'll need to port the 2 config files to 56-branch (in case we'll use for MC the same tag as for raw reco re-filtering).

Are the pass1 MC productions affected by the same issue? Probably not, as it would have been spotted already by now. Just in case, for pass1 MC I think I remember some productions where we had [0, 6.5] fm impact parameter range, which would need to update to bmax<=6.5 - but I'm not sure if this is even the case, so we should be fine already now. We're not going to do much anyway for pass1 productions.

Thanks a lot!

Yours,
Catalin.

fprino:
Hi Catalin,

from what reported by Fabrizio, the problem was there also in the pass1 MCs, even though with reduced strength.
On the refilterigns a consideration came into my mind: for most (if not all) the HF MC productions we do not keep the ESDs, so the issue of running a refiltering will probably never occur (this holds in particular for AOD242).

Cheers, Francesco

PR #447, opened 2020-11-25
Add missing line to retrieve environment variable

fprino:
This is the fix needed for ALIROOT-8517, as reported by Catalin in the Jira ticket


mpuccio:
Thanks and sorry I missed that!

PR #455, opened 2021-02-01
AOD macro: removed filter trees merging and corrected AOD PS EventSta…

catalinristea:
…t_temp_AODFiltering

### DIFF
--- a/AOD/main_AODtrainRawAndMC.C
+++ b/AOD/main_AODtrainRawAndMC.C
@@ -902,7 +902,7 @@ void AODmerge()
   TStopwatch timer;
   timer.Start();
   TString outputDir = "wn.xml";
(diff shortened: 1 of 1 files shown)


catalinristea:
Dear all,

This update has been tested on both central RAW data reco and MC AOD mergings:

1. It removes the FilterTrees from the AOD merging, as the trees are much less fragmented (due to their size, compared with the limitation in larger AOD size) in a separate job - this allows to spare some storage space;
2. Due to the fact we're running physics selection in all 3 macros (QA, AOD and QAtrainAOD), we produce (probably identical, at least in file size) EventStat_temp (QA), EventStat_temp_AOD (QAtrainAOD) and EventStat_temp_AODFiltering (AOD) - for AOD merging we were in fact merging the QA EventStat_temp, but now we are updating to use the correct EventStat_temp_AODFiltering (just for the sake of argument, not that this would matter);
3. This has not been checked yet on re-filtering, but it shouldn't do any harm and if it does, then we'll correct at a later point;

Hence I'll merge this pull request.

Thanks and kind regards,
Catalin.

PR #456, opened 2021-02-16
Fix STARLight parameter setting for older AliRoot versions

miweberSMI:


miweberSMI:
Hi @catalinristea, @gconesab,
with this fix the latest 2015 LMee production (ALIROOT-8634) should work (I am still testing it), feel free to either merge it and have a new tag or a custom GeneratorConfig.C file.
Cheers,
Michael

PR #467, opened 2021-05-12
Add case of simulation/reconstruction of EMCal plus trigger detectors

gconesab:


catalinristea:
Thanks lot, Gustavo!

Yours,
Catalin.

PR #468, opened 2021-05-17
Add generator for absorption studies

mpuccio:
This new generator will be needed for an upcoming MC request.

@mcolocci please check the pt ranges of the generators

### DIFF
--- a/MC/CustomGenerators/PWGLF/Hijing_AbsStudies.C
+++ b/MC/CustomGenerators/PWGLF/Hijing_AbsStudies.C
@@ -1,0 +1,22 @@
+AliGenerator *
+GeneratorCustom(TString opt = "d")
+{
+    TString optList[4] = {"a", "b", "c", "d"};
+    Int_t ninjH3[4] = {25, 15, 10, 5};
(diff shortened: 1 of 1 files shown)


mcolocci:
> 
> 
> This new generator will be needed for an upcoming MC request.
> 
> @mcolocci please check the pt ranges of the generators

Many thanks @mpuccio for doing this.
Fine with p = [0.5, 3] GeV/c for (anti)tritons.
Instead I would go lower for He-3.
That is, I would choose p = [1, 10] GeV/c for He-3.
If we go a bit higher in momentum is also fine.
Many thanks again.

mpuccio:
Ciao Manuel,
but don't we have a cut at 3 GeV/c minimum for He3? I put 2GeV/c for safety, but I am not sure what do we gain going lower.
Cheers,
Max

mcolocci:
Yes. However by using the pass3 (not the special reconstruction) we can go lower in momentum.
Then, we will need to cut "offline" at 3 cm in TOF (in the MC as well).
We saw indeed that there is no large extra signal out of that.

mpuccio:
OK, updated! I will open the ticket this evening

PR #475, opened 2021-09-02
Enable FOCAL simulations with alidpg

iarsene:
Files touched:
1 MC/dpgsim.sh
1.1 Support for the "--detector FOCAL" option, which enables a special list of detector (see DetConfig.C)
1.2 Added the option "--focalGeometryFile " to be able to switch easily between different geometries
1.3 Only the "sim" mode is available for FOCAL simulations with these changes. No FOCAL digitization yet!!

2 MC/Config_LoadLibraries.C
Load the FOCAL libraries from the FOCAL external package if "--detector FOCAL" option is present

3 MC/DetectorConfig.C
3.1 Implemented an EDetector_t option for FOCAL and enabled/disabled needed detectors.
3.2 In DetectorInit(), run the AliPIPEFOCAL and FOCAL constructors for FOCAL simulations

4 MC/sim.C
Switch off LoadAlignFromCDB() and UseDetectorsFromGRP() for the FOCAL case, since there is no FOCAL alignment entry in the CDB yet.

Added MC/CustomGenerators/Upgrade/FOCAL_Generators.C
This contains several tailored generators for FOCAL simulations

### DIFF
--- a/MC/Config_LoadLibraries.C
+++ b/MC/Config_LoadLibraries.C
@@ -36,6 +36,14 @@ Config_LoadLibraries()
   } 
   TString isFluka = gSystem->Getenv("CONFIG_FLUKA");
   if (isFluka!="on") gSystem->Load("libgeant321");
-
+  
+  if (gSystem->Getenv("CONFIG_DETECTOR")) {
+    if (strcmp(gSystem->Getenv("CONFIG_DETECTOR"), "FOCAL") == 0) {
+      printf(">>>>> Load FOCAL libs \n");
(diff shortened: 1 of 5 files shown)


gconesab:
It looks good to me, but if it is not too hard, I would prefer to avoid the indentation changes in dpgsim.sh so that we see what was actually changed

iarsene:
I reversed the indentation, but git still sees differences because of spaces used instead of tabs i think. I will checkout the old file and apply just my changes.

iarsene:
Indentation removed

PR #478, opened 2021-09-22
Updated injection scheme

alcaliva:
Injecting 20 nuclei per species instead of 10.


gconesab:
There is a conflict I do not understand. Can you close and restart the changes after rebasing?

PR #487, opened 2021-11-03
Bug fix 

alcaliva:


gconesab:
I think you need to open an new PR, there are conflicts. 
Why do you need to put the generator into a cocktail?
Gustavo

alcaliva:
I would need to generate pythia8 events with a trigger particle having a pt>5 GeV. 
Is there a better way?

gconesab:
In the same way you where doing before this commit. I do not understand the new approach. Maybe just you need the casting to AliGenPythiaPlus.

Gustavo

alcaliva:
OK, will try. Thanks

PR #495, opened 2021-12-06
Update the max pt of the lambda @maciacco

mpuccio:


mpuccio:
Hi @gconesab,
yes it is intended, the Lambda have an extended pt range because we need it also for the B_Lambda measurement.
Cheers,
Max

PR #507, opened 2022-03-04
add option to set ITS material budget scaling factors

noferini:


noferini:
@chiarazampolli @catalinristea @gconesab 
Can you check that is fine with you?
To be tested with simulations
Cheers
Francesco

chiarazampolli:
Thanks!! Seems fine to me, but I have not tested it. Merging, since it won't break anything.

PR #517, opened 2022-08-09
Fixed issue in uid use for xi PDG code sign

bianchil:


chiarazampolli:
Hi @bianchil , 

thanks! To be sure, did you run a small production to verify that the configuration is ok before we move to the grid?

Chiara

bianchil:
Hi @chiarazampolli ,
yes of course. I always do :) Problem last time was that I tested at fixed UID (1), while the problem was with even UIDs. Now tested with both.
Sorry for the inconvenience.
Livio

chiarazampolli:
Thanks @bianchil !

